PR #1, opened 2019-10-08
Fix tag detection issue for image replacement in k8s-deploy

Publiskills:
This is the fix for the issue https://github.com/Azure/k8s-deploy/issues

### DIFF
--- a/lib/kubernetes-utils.js
+++ b/lib/kubernetes-utils.js
@@ -97,7 +97,10 @@ function substituteImageNameInSpecContent(currentString, imageName, imageNameWit
 function updateContainerImagesInManifestFiles(contents, containers) {
     if (!!containers && containers.length > 0) {
         containers.forEach((container) => {
-            let imageName = container.split(':')[0];
+            let imageName = container;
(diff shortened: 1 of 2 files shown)


lib/kubernetes-utils.js
```diff
@@ -80,13 +80,19 @@ function substituteImageNameInSpecContent(currentString, imageName, imageNameWit
...
+
+            const registrySplitIndex = currentImageName.indexOf('/');
+            const lastTagSplitIndex = currentImageName.lastIndexOf(':');
+            if(lastTagSplitIndex > registrySplitIndex) {
+                currentImageName = currentImageName.substring(0, lastTagSplitIndex)
+            }
+
             if (currentImageName === imageName) {
                 return acc + `${imageKeyword[0]} ${imageNameWithNewTag}\n`;
             }
```
ds-ms: Instead of this, you could consider parsing the yaml to `json` and using `jsonpath` you could update the image fields

ds-ms:
Raised a PR for handling different issues related to the replacement logic here #13

Issue #10, opened 2019-11-27
Deployment/update error map: map[] does not contain declared merge key: name


niekappartmedia:
When i deploy the first time with k8s deploy everything works fine.
But when i deploy the second time i get the following error:

error calculating patch from openapi spec: map: map[] does not contain declared merge key: name

Strange though that when i deploy localy the firsttime with kubectl and update everything works fine.
But when k8s has deployed i even can not update with local kubectl i get the same error.

Any ideas what this could be?

This is the deployment yaml:

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test
  labels:
    app: test
spec:
  replicas: 1
  selector:
    matchLabels:
      app: test
  strategy:
    type: RollingUpdate
  template:
    metadata:
      name: test
      labels:
        app: test
    spec:
      securityContext:
        runAsGroup: 33
        fsGroup: 33
      volumes:
        - name: nfs-bucket
          persistentVolumeClaim:
            claimName: nfs-bucket
      containers:
        - name: test
          image: niek/test:develop
          imagePullPolicy: Always
          volumeMounts:
            - name: nfs-bucket
              mountPath: "/var/www/bucket"
          envFrom:
            - secretRef:
                name: test-secret
          ports:
            - name: 80tcp01
              containerPort: 80
              protocol: TCP
          stdin: true
          tty: true
      imagePullSecrets:
        - name: dockerhub
```

niekappartmedia:
So i am a bit further with this issue.

After testing and investigating.
When i update via kubectl -f it works when i comment the imagePullSecret.
After this step when i update again with imagePullSecret uncommented it works also also for azure/k8s-deploy. But after that when i again want to redeploy via azure/k8s-deploy i get the same error. So it looks like the imagePullSecrets are causing all te trouble. Not yet found the fix for this.

lennartschoch:
I am also experiencing this, but in a slightly different scenario:
I do not specify any imagePullSecrets, yet I get the same error.

What I thought is that kubectl was doing some weird transformation of the YAML file, causing it to map `"imagePullSecrets":[ "name":""]` (which seems to be the value that azure/k8s-deploy passes to kubectl) to `"imagePullSecrets":[map[]]` which is invalid because of the missing "name" key.

I am not sure why this is coming up right now, as azure/k8s-deploy did not change. Maybe it has to do something with the kubectl version. I will continue to look into this

niekappartmedia:
@lennartschoch 
[...]

PR #12, opened 2019-11-29
issue #10 fix missing name key

lennartschoch:
Added non-empty validation to pullImageSecrets. This should fix issue #10.


ds-ms:
LGTM

ds-ms:
Could you raise the same PR for master as well

lennartschoch:
Will do so :)

PR #16, opened 2020-01-09
Users/shigupt/traffic split api version update

shigupt202:


lib/utilities/strategy-helpers/smi-canary-deployment-helper.js
```diff
@@ -8,10 +8,12 @@ const TaskInputParameters = require("../../input-parameters");
 const fileHelper = require("../files-helper");
 const helper = require("../resource-object-utility");
 const utils = require("../manifest-utilities");
+const kubectlUtils = require("../../kubectl-util");
 const canaryDeploymentHelper = require("./canary-deployment-helper");
 const utility_1 = require("../utility");
 const TRAFFIC_SPLIT_OBJECT_NAME_SUFFIX = '-workflow-rollout';
 const TRAFFIC_SPLIT_OBJECT = 'TrafficSplit';
+var trafficSplitAPIVersion = null;
```
ds-ms: Use `let` instead

ds-ms: Set it to empty string or a set a default value instead of null

lib/utilities/strategy-helpers/smi-canary-deployment-helper.js
```diff
@@ -145,24 +147,27 @@ function adjustTraffic(kubectl, manifestFilePaths, stableWeight, canaryWeight) {
...
+function createTrafficSplitManifestFile(kubectl, serviceName, stableWeight, baselineWeight, canaryWeight) {
+    const smiObjectString = getTrafficSplitObject(kubectl, serviceName, stableWeight, baselineWeight, canaryWeight);
     const manifestFile = fileHelper.writeManifestToFile(smiObjectString, TRAFFIC_SPLIT_OBJECT, serviceName);
     if (!manifestFile) {
         throw new Error('UnableToCreateTrafficSplitManifestFile');
     }
     return manifestFile;
 }
-function getTrafficSplitObject(name, stableWeight, baselineWeight, canaryWeight) {
+function getTrafficSplitObject(kubectl, name, stableWeight, baselineWeight, canaryWeight) {
```
ds-ms: Add an extra line between functions

src/utilities/strategy-helpers/smi-canary-deployment-helper.ts
```diff
@@ -10,11 +10,13 @@ import * as TaskInputParameters from '../../input-parameters';
 import * as fileHelper from '../files-helper';
 import * as helper from '../resource-object-utility';
 import * as utils from '../manifest-utilities';
+import * as kubectlUtils from '../../kubectl-util';
```
ds-ms: Why's the file two directories up? 
Add it to utilities folder

src/utilities/strategy-helpers/smi-canary-deployment-helper.ts
```diff
@@ -10,11 +10,13 @@ import * as TaskInputParameters from '../../input-parameters';
 import * as fileHelper from '../files-helper';
 import * as helper from '../resource-object-utility';
 import * as utils from '../manifest-utilities';
+import * as kubectlUtils from '../../kubectl-util';
 import * as canaryDeploymentHelper from './canary-deployment-helper';
 import { checkForErrors } from "../utility";
 
 const TRAFFIC_SPLIT_OBJECT_NAME_SUFFIX = '-workflow-rollout';
 const TRAFFIC_SPLIT_OBJECT = 'TrafficSplit';
+var trafficSplitAPIVersion = null;
```
ds-ms: Same as above

ds-ms: Address the comments in the `js` file. I missed seeing the extension of that file

lib/utilities/strategy-helpers/smi-canary-deployment-helper.js
shigupt202: js files are automatically generated during build and there's no extra line between any of the functions in any file. Should I still do it for this function?

lib/utilities/strategy-helpers/smi-canary-deployment-helper.js
shigupt202: Okay

lib/utilities/strategy-helpers/smi-canary-deployment-helper.js
ds-ms: No no. I mistook it for a ts file; you can ignore this comment

src/utilities/strategy-helpers/smi-canary-deployment-helper.ts
shigupt202: It was in the same folder as run.ts and manifest-utilities.ts, because these were the only two files using kubectl-utils previously. Since smi-canary-deployment-helper.ts is also using it, I'll move it to utilities now.

src/utilities/kubectl-util.ts
```diff
@@ -1,65 +1,76 @@
...
+    }
+
+    const kubectlPath = path.join(cachedToolpath, kubectlToolName + getExecutableExtension());
+    fs.chmodSync(kubectlPath, '777');
+    return kubectlPath;
+}
+
+export function getTrafficSplitAPIVersion(kubectl: Kubectl): string {
+    const result = kubectl.executeCommand('api-versions');
+    const trafficSplitAPIVersion = result.stdout.split('\n').find(version => version.startsWith('split.smi-spec.io'));
```
ammohant: Can you please explore making "split.smi-spec.io" as a constant and use the constant here ?

src/utilities/strategy-helpers/smi-canary-deployment-helper.ts
```diff
@@ -90,7 +92,7 @@ function createCanaryService(kubectl: Kubectl, filePaths: string[]) {
                     newObjectsList.push(newStableServiceObject);
 
                     core.debug('Creating the traffic object for service: ' + name);
-                    const trafficObject = createTrafficSplitManifestFile(name, 0, 0, 1000);
+                    const trafficObject = createTrafficSplitManifestFile(kubectl, name, 0, 0, 1000);
```
ammohant: can we keep the constants in CAPs.

shigupt202: This is just a variable name. I think we can keep it like this, just like other consts.

src/utilities/kubectl-util.ts
shigupt202: Okay

src/kubectl-object-model.ts
```diff
@@ -79,6 +79,9 @@ export class Kubectl {
         return results;
     }
 
+    public executeCommand(customCommand: string, args?: string) {
+        return args ? this.execute([customCommand, args]) : this.execute([customCommand]);
```
ds-ms: Add null check for `customCommand` as well

src/utilities/kubectl-util.ts
```diff
@@ -1,65 +1,77 @@
...
+
+    const kubectlPath = path.join(cachedToolpath, kubectlToolName + getExecutableExtension());
+    fs.chmodSync(kubectlPath, '777');
+    return kubectlPath;
+}
+
+export function getTrafficSplitAPIVersion(kubectl: Kubectl): string {
+    const result = kubectl.executeCommand('api-versions');
+    const trafficSplitAPIVersion = result.stdout.split('\n').find(version => version.startsWith(trafficSplitAPIVersionPrefix));
+    if (trafficSplitAPIVersion == null || typeof trafficSplitAPIVersion == 'undefined') {
```
ds-ms: `if (!trafficSplitAPIVersion)` should do given that `undefined` & `null` are equivalent to false in js

src/kubectl-object-model.ts
shigupt202: Should I throw an error when no command is given, or should I just return from the function with null?

ds-ms: What's the current usecase for this function? Do you want to break it when the command is set to an empty string?

shigupt202: The function is currently being used to execute kubectl api-versions command only. I think it makes more sense to throw an error in case of empty command string.

ds-ms: Makes sense to throw it then :)

PR #17, opened 2020-01-10
Users/shigupt/adding kubectl describe output in logs

shigupt202:


src/utilities/manifest-stability-utility.ts
```diff
@@ -59,15 +70,19 @@ export async function checkPodStatus(kubectl: Kubectl, podName: string): Promise
...
             break;
         case 'Pending':
             if (!isPodReady(podStatus)) {
                 core.warning(`pod/${podName} rollout status check timedout`);
+                kubectl.describe('pod', podName);
             }
             break;
         case 'Failed':
             core.error(`pod/${podName} rollout failed`);
+            kubectl.describe('pod', podName);
```
ds-ms: The call during errors seems to be spread out. Try to call it at one point, will come in handy if/when you want to add some extra stuff later

shigupt202: Should I create a function and put describe command in it?

ds-ms: Set a variable for indicator and do the describe call in one place

PR #31, opened 2020-05-06
Modified the description to optimise marketplace search

anraghun:
#### Contents
- Modification to the description of the action 💬

#### Purpose
To optimise the marketplace search for this action ✌️


anraghun:
@N-Usha Can you review this one as well?
@ds-ms Will need collab access on this one too.

I guess I'll need to port this as well to `releases/v1` branch?

action.yml
```diff
@@ -1,5 +1,5 @@
 name: 'Deploy to Kubernetes cluster'
-description: 'Deploy to Kubernetes cluster'
+description: 'Deploy to a Kubernetes cluster such as but not limited to Azure Kubernetes Service clusters'
```
ds-ms, line 2: `including, but not limited to` maybe?

anraghun: Yup, sounds cleaner.

PR #35, opened 2020-05-28
Test Cases for deploy

rgsubh:
fixes #28

PR #36, opened 2020-06-01
Test Cases for deploy (#35)

rgsubh:
* Test Cases for deploy

* Testing workflow

* Fixing an issue with test runs on remote

* Fixing an issue with test runs on remote

* Fixing an issue with test runs on remote

* Fixing an issue with test runs on remote

* Fixing an issue with test runs on remote

* Fixing an issue with test runs on remote

* Review Comments fix

* Review comments fix

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -1,25 +1,35 @@
 'use strict';
 Object.defineProperty(exports, "__esModule", { value: true });
-class KubernetesWorkload {
-}
-KubernetesWorkload.pod = 'Pod';
-KubernetesWorkload.replicaset = 'Replicaset';
-KubernetesWorkload.deployment = 'Deployment';
-KubernetesWorkload.statefulSet = 'StatefulSet';
-KubernetesWorkload.daemonSet = 'DaemonSet';
-KubernetesWorkload.job = 'job';
-KubernetesWorkload.cronjob = 'cronjob';
+exports.workloadTypesWithRolloutStatus = exports.workloadTypes = exports.deploymentTypes = exports.ServiceTypes = exports.DiscoveryAndLoadBalancerResource = exports.KubernetesWorkload = void 0;
+let KubernetesWorkload = /** @class */ (() => {
+    class KubernetesWorkload {
+    }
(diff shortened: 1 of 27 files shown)


lib/constants.js
```diff
@@ -1,25 +1,35 @@
...
-class KubernetesWorkload {
-}
-KubernetesWorkload.pod = 'Pod';
-KubernetesWorkload.replicaset = 'Replicaset';
-KubernetesWorkload.deployment = 'Deployment';
-KubernetesWorkload.statefulSet = 'StatefulSet';
-KubernetesWorkload.daemonSet = 'DaemonSet';
-KubernetesWorkload.job = 'job';
-KubernetesWorkload.cronjob = 'cronjob';
+exports.workloadTypesWithRolloutStatus = exports.workloadTypes = exports.deploymentTypes = exports.ServiceTypes = exports.DiscoveryAndLoadBalancerResource = exports.KubernetesWorkload = void 0;
```
ds-ms, line 3: Did you make any changes to `constants.ts` file? Or this convention was changed due to a compiler version diff?

src/run.ts
```diff
@@ -1,14 +1,15 @@
-import * as toolCache from '@actions/tool-cache';
 import * as core from '@actions/core';
 import * as io from '@actions/io';
 import * as path from 'path';
+import * as toolCache from '@actions/tool-cache';
```
ds-ms, line 4: Why the reordering?

src/run.ts
```diff
@@ -57,7 +58,7 @@ async function run() {
         namespace = 'default';
     }
     let action = core.getInput('action');
-    let manifests = manifestsInput.split('\n');
+    let manifests = manifestsInput ? manifestsInput.split('\n') : undefined;
```
ds-ms, line 61: Shouldn't it be `""` or `[]` instead?

ds-ms: You'll end up seeing some type mismatch down the line.

ds-ms: `!manifestsInput` would never be hit due to the check in line 53. You probably should just add a `return` statement to break the action flow there.

Issue #24, opened 2020-02-25
Add support for an optional flag --force


s-tuli:
Add support for an optional flag --force

Our team's (Azure dev Spaces) customer scenario requires us to deploy when a previous deployment already exists. It cannot be a clean deployment. And in this case, I get an error "file is immutable".

Please add it as an optional flag whose value is false by default. We will add a value true as part of our workflow.

**Additional Detail**: If a job has been created already, you cannot replace without using --force. 
One cannot replace jobs since Kubernetes adds labels on create. This is by design.
Specifically this error occurs for: https://github.com/Azure/dev-spaces/blob/master/samples/BikeSharingApp/PopulateDatabase/charts/populatedatabase/templates/job.yaml

Thanks!

ds-ms:
@s-tuli Could you explain your scenario in detail, for building context in this thread.

s-tuli:
> @s-tuli Could you explain your scenario in detail, for building context in this thread.

Added additional context above @ds-ms

ds-ms:
@s-tuli I meant your use-case.

ds-ms:
@tauhid621 to take a look

PR #43, opened 2020-06-15
optional force flag added

tauhid621:
Optional force flag added. If force flag is on and patch fails, then resource will be deleted and re-created.
#24


lib/actions/promote.js
```diff
@@ -1,51 +1,52 @@
...
-        try {
-            canaryDeploymentHelper.deleteCanaryDeployment(kubectl, TaskInputParameters.manifests, includeServices);
-        }
-        catch (ex) {
-            core.warning('Exception occurred while deleting canary and baseline workloads. Exception: ' + ex);
-        }
-    });
-}
-exports.promote = promote;
+'use strict';
```
ds-ms, line 1: @tauhid621 could you undo your LF to CRLF changes

src/run.ts
```diff
@@ -61,10 +61,12 @@ export async function run() {
     let action = core.getInput('action');
     let manifests = manifestsInput.split('\n');
 
+    let forceDeployment: boolean = core.getInput('force') === 'true';
+
     if (action === 'deploy') {
         let strategy = core.getInput('strategy');
         console.log("strategy: ", strategy)
-        await deploy(new Kubectl(kubectlPath, namespace), manifests, strategy);
+        await deploy(new Kubectl(kubectlPath, namespace, undefined, forceDeployment), manifests, strategy);
```
ds-ms, line 69: Send `false` instead of `undefined`

ds-ms: Is it being set to `true` anywhere btw?

shigupt202:
Can you update the tests for the changes that you've made? The CI checks are failing.

src/kubectl-object-model.ts
```diff
@@ -21,7 +23,13 @@ export class Kubectl {
...
 
     public apply(configurationPaths: string | string[]) {
-        return this.execute(['apply', '-f', this.createInlineArray(configurationPaths)]);
+        if (this.forceDeployment) {
+            console.log("force flag is on, deployment will continue even if previous deployment already exists");
+            return this.execute(['apply', '--force', '-f', this.createInlineArray(configurationPaths)]);
+        }
+        else {
+            return this.execute(['apply', '-f', this.createInlineArray(configurationPaths)]);
+        }
```
ds-ms, line 32: Call execute from one place.

Do something like
```
initiate array
if(force) add --force to array
array.append(more args)

this.execute()
```

Don't you think this looks cleaner

src/kubectl-object-model.ts
```diff
@@ -9,10 +9,12 @@ export class Kubectl {
     private kubectlPath: string;
     private namespace: string;
     private ignoreSSLErrors: boolean;
+    private forceDeployment: boolean;
 
-    constructor(kubectlPath: string, namespace?: string, ignoreSSLErrors?: boolean) {
+    constructor(kubectlPath: string, namespace?: string, ignoreSSLErrors?: boolean, forceDeployment?: boolean) {
```
ds-ms, line 14: Also, given that this is only to be used for `apply` doesn't it make more sense to be a parameter to that function instead?

PR #44, opened 2020-06-16
Optional force flag added

tauhid621:
Optional Force flag added. If true, this will immediately remove resources from API and bypass graceful deletion.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -41,6 +41,10 @@ inputs:
     description: 'deploy/promote/reject'
     required: true
     default: 'deploy'
+  force:
+    description: 'Deploy when a previous deployment already exists'
+    required: false
+    default: false
 
 branding:
   color: 'green' # optional, decorates the entry in the GitHub Marketplace
--- a/action.yml
+++ b/action.yml
@@ -42,7 +42,7 @@ inputs:
     required: true
     default: 'deploy'
   force:
-    description: 'Deploy when a previous deployment already exists. If true then --force argument is added to the apply command'.
+    description: 'Deploy when a previous deployment already exists. If true then --force argument is added to the apply command.'
     required: false
     default: false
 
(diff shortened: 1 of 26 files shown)


lib/actions/reject.js
```diff
@@ -9,6 +9,7 @@ var __awaiter = (this && this.__awaiter) || function (thisArg, _arguments, P, ge
     });
 };
 Object.defineProperty(exports, "__esModule", { value: true });
+exports.reject = void 0;
```
ds-ms, line 12: Make sure you update the `typescript` version in the package.json file to avoid this later

src/kubectl-object-model.ts
```diff
@@ -21,7 +22,14 @@ export class Kubectl {
     }
 
     public apply(configurationPaths: string | string[]) {
-        return this.execute(['apply', '-f', this.createInlineArray(configurationPaths)]);
+        let applyArgs: string[] = ['apply', '-f', this.createInlineArray(configurationPaths)];
+
+        if (TaskInputParameters.forceDeployment) {
+            console.log("force flag is on, deployment will continue even if previous deployment already exists");
+            applyArgs.push('--force');
```
ds-ms, line 29: Does the position of `--force` flag impact the outcome?

ds-ms:
NIT: the branch names don't need to start with `users/*`.

ds-ms:
Add changes to README.md file as well. 

Do mention how using `--force` is not recommended in production

action.yml
```diff
@@ -41,6 +41,10 @@ inputs:
     description: 'deploy/promote/reject'
     required: true
     default: 'deploy'
+  force:
+    description: 'Deploy when a previous deployment already exists'
```
ds-ms, line 45: Add something like.. `adds `--force` argument to the apply command` or something on those lines?

tauhid621: added

PR #47, opened 2020-06-22
added blue green strategy

sundargs2000:
Like the existing canary deployment strategy, added blue-green deployment strategy to the actions with three route-methods - service, ingress and smi. also added version-switch-buffer input takes input in minutes. It waits for the given time before routing service/ingress/trafficsplit to new deployments in deploy step. 

Reference image for how service routing should work.
service-diagram

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -19,9 +19,17 @@ inputs:
       description: 'Version of kubectl. Installs a specific version of kubectl binary'
       required: false
   strategy:
-    description: 'Deployment strategy to be used. Allowed values are none, canary'
+    description: 'Deployment strategy to be used. Allowed values are none, canary and blue-green'
     required: false
     default: 'none'
+  route-method:
+    description: 'Route based on service, ingress or SMI for blue-green strategy'
+    required: false
+    default: 'service'
+  version-switch-buffer:
+    description: 'Indicates the buffer time in minutes before the switch is made to the green version (max is 300 min ie. 5hrs)'
+    required: false
+    default: 0
   traffic-split-method:
     description: "Traffic split method to be used. Allowed values are pod, smi"
     required: false
--- a/src/actions/promote.ts
+++ b/src/actions/promote.ts
@@ -1,22 +1,35 @@
 'use strict';
 import * as core from '@actions/core';
-
 import * as deploymentHelper from '../utilities/strategy-helpers/deployment-helper';
 import * as canaryDeploymentHelper from '../utilities/strategy-helpers/canary-deployment-helper';
 import * as SMICanaryDeploymentHelper from '../utilities/strategy-helpers/smi-canary-deployment-helper';
 import * as utils from '../utilities/manifest-utilities';
 import * as TaskInputParameters from '../input-parameters';
+import { getUpdatedManifestFiles } from '../utilities/manifest-utilities'
+import * as KubernetesObjectUtility from '../utilities/resource-object-utility';
+import * as models from '../constants';
+import * as KubernetesManifestUtility from '../utilities/manifest-stability-utility';
+import { getManifestObjects, deleteWorkloadsWithLabel, deleteWorkloadsAndServicesWithLabel } from '../utilities/strategy-helpers/blue-green-helper';
+import { BLUE_GREEN_NEW_LABEL_VALUE, NONE_LABEL_VALUE } from '../utilities/strategy-helpers/blue-green-helper';
+import { isBlueGreenDeploymentStrategy, blueGreenRouteService, blueGreenPromote } from '../utilities/strategy-helpers/service-blue-green-helper';
+import { isIngressRoute, blueGreenRouteIngress, blueGreenPromoteIngress } from '../utilities/strategy-helpers/ingress-blue-green-helper';
+import { isSMIRoute, blueGreenRouteTraffic, blueGreenPromoteSMI, cleanSetUpSMI } from '../utilities/strategy-helpers/smi-blue-green-helper';
+import { Kubectl, Resource } from '../kubectl-object-model';
 
-import { Kubectl } from '../kubectl-object-model';
+export async function promote() {
+    const kubectl = new Kubectl(await utils.getKubectl(), TaskInputParameters.namespace, true);
 
-export async function promote(ignoreSslErrors?: boolean) {
-    const kubectl = new Kubectl(await utils.getKubectl(), TaskInputParameters.namespace, ignoreSslErrors);
-
-    if (!canaryDeploymentHelper.isCanaryDeploymentStrategy()) {
-        core.debug('Strategy is not canary deployment. Invalid request.');
+    if (canaryDeploymentHelper.isCanaryDeploymentStrategy()) {
+        await promoteCanary(kubectl);
+    } else if (isBlueGreenDeploymentStrategy()) {
+        await promoteBlueGreen(kubectl);
+    } else {
+        core.debug('Strategy is not canary or blue-green deployment. Invalid request.');
         throw ('InvalidPromotetActionDeploymentStrategy');
     }
+}
 
+async function promoteCanary(kubectl: Kubectl) {
     let includeServices = false;
     if (canaryDeploymentHelper.isSMICanaryStrategy()) {
         includeServices = true;
@@ -40,5 +53,40 @@ export async function promote(ignoreSslErrors?: boolean) {
         canaryDeploymentHelper.deleteCanaryDeployment(kubectl, TaskInputParameters.manifests, includeServices);
     } catch (ex) {
         core.warning('Exception occurred while deleting canary and baseline workloads. Exception: ' + ex);
+    }
+}
+
+async function promoteBlueGreen(kubectl: Kubectl) {
+    // updated container images and pull secrets
+    let inputManifestFiles: string[] = getUpdatedManifestFiles(TaskInputParameters.manifests);
+    const manifestObjects = getManifestObjects(inputManifestFiles);
+
+    core.debug('deleting old deployment and making new ones');
+    let result;
+    if(isIngressRoute()) {
+        result = await blueGreenPromoteIngress(kubectl, manifestObjects);
+    } else if (isSMIRoute()) {
+        result = await blueGreenPromoteSMI(kubectl, manifestObjects);
+    } else {
+        result = await blueGreenPromote(kubectl, manifestObjects);
+    }
+
+    // checking stability of newly created deployments 
+    const deployedManifestFiles = result.newFilePaths;
+    const resourceTypes: Resource[] = KubernetesObjectUtility.getResources(deployedManifestFiles, models.deploymentTypes.concat([models.DiscoveryAndLoadBalancerResource.service]));
+    await KubernetesManifestUtility.checkManifestStability(kubectl, resourceTypes);
+    
+    core.debug('routing to new deployments');
+    await KubernetesManifestUtility.checkManifestStability(kubectl, resourceTypes);
+    if(isIngressRoute()) {
+        blueGreenRouteIngress(kubectl, null, manifestObjects.serviceNameMap, manifestObjects.serviceEntityList, manifestObjects.ingressEntityList);
+        deleteWorkloadsAndServicesWithLabel(kubectl, BLUE_GREEN_NEW_LABEL_VALUE, manifestObjects.deploymentEntityList, manifestObjects.serviceEntityList);
+    } else if (isSMIRoute()) {
+        blueGreenRouteTraffic(kubectl, NONE_LABEL_VALUE, manifestObjects.deploymentEntityList, manifestObjects.serviceEntityList);
+        deleteWorkloadsWithLabel(kubectl, BLUE_GREEN_NEW_LABEL_VALUE, manifestObjects.deploymentEntityList);
+        cleanSetUpSMI(kubectl, manifestObjects.deploymentEntityList, manifestObjects.serviceEntityList);    
+    } else {
+        blueGreenRouteService(kubectl, NONE_LABEL_VALUE, manifestObjects.deploymentEntityList, manifestObjects.serviceEntityList);
+        deleteWorkloadsWithLabel(kubectl, BLUE_GREEN_NEW_LABEL_VALUE, manifestObjects.deploymentEntityList);
     }
 }
--- a/src/actions/promote.ts
+++ b/src/actions/promote.ts
@@ -10,10 +10,10 @@ import * as KubernetesObjectUtility from '../utilities/resource-object-utility';
 import * as models from '../constants';
 import * as KubernetesManifestUtility from '../utilities/manifest-stability-utility';
 import { getManifestObjects, deleteWorkloadsWithLabel, deleteWorkloadsAndServicesWithLabel } from '../utilities/strategy-helpers/blue-green-helper';
-import { isBlueGreenDeploymentStrategy, isIngressRoute, isSMIRoute, BLUE_GREEN_NEW_LABEL_VALUE, NONE_LABEL_VALUE } from '../utilities/strategy-helpers/blue-green-helper';
-import { routeBlueGreenService, blueGreenPromote } from '../utilities/strategy-helpers/service-blue-green-helper';
-import { routeBlueGreenIngress, blueGreenPromoteIngress } from '../utilities/strategy-helpers/ingress-blue-green-helper';
-import { routeBlueGreenSMI, blueGreenPromoteSMI, cleanupSMI } from '../utilities/strategy-helpers/smi-blue-green-helper';
+import { isBlueGreenDeploymentStrategy, isIngressRoute, isSMIRoute, GREEN_LABEL_VALUE, NONE_LABEL_VALUE } from '../utilities/strategy-helpers/blue-green-helper';
+import { routeBlueGreenService, promoteBlueGreenService } from '../utilities/strategy-helpers/service-blue-green-helper';
+import { routeBlueGreenIngress, promoteBlueGreenIngress } from '../utilities/strategy-helpers/ingress-blue-green-helper';
+import { routeBlueGreenSMI, promoteBlueGreenSMI, cleanupSMI } from '../utilities/strategy-helpers/smi-blue-green-helper';
 import { Kubectl, Resource } from '../kubectl-object-model';
 
 export async function promote() {
@@ -64,29 +64,28 @@ async function promoteBlueGreen(kubectl: Kubectl) {
     core.debug('deleting old deployment and making new ones');
     let result;
     if(isIngressRoute()) {
-        result = await blueGreenPromoteIngress(kubectl, manifestObjects);
+        result = await promoteBlueGreenIngress(kubectl, manifestObjects);
     } else if (isSMIRoute()) {
-        result = await blueGreenPromoteSMI(kubectl, manifestObjects);
+        result = await promoteBlueGreenSMI(kubectl, manifestObjects);
     } else {
-        result = await blueGreenPromote(kubectl, manifestObjects);
+        result = await promoteBlueGreenService(kubectl, manifestObjects);
     }
 
     // checking stability of newly created deployments 
     const deployedManifestFiles = result.newFilePaths;
-    const resourceTypes: Resource[] = KubernetesObjectUtility.getResources(deployedManifestFiles, models.deploymentTypes.concat([models.DiscoveryAndLoadBalancerResource.service]));
-    await KubernetesManifestUtility.checkManifestStability(kubectl, resourceTypes);
+    const resources: Resource[] = KubernetesObjectUtility.getResources(deployedManifestFiles, models.deploymentTypes.concat([models.DiscoveryAndLoadBalancerResource.service]));
+    await KubernetesManifestUtility.checkManifestStability(kubectl, resources);
     
     core.debug('routing to new deployments');
-    await KubernetesManifestUtility.checkManifestStability(kubectl, resourceTypes);
     if(isIngressRoute()) {
         routeBlueGreenIngress(kubectl, null, manifestObjects.serviceNameMap, manifestObjects.serviceEntityList, manifestObjects.ingressEntityList);
-        deleteWorkloadsAndServicesWithLabel(kubectl, BLUE_GREEN_NEW_LABEL_VALUE, manifestObjects.deploymentEntityList, manifestObjects.serviceEntityList);
+        deleteWorkloadsAndServicesWithLabel(kubectl, GREEN_LABEL_VALUE, manifestObjects.deploymentEntityList, manifestObjects.serviceEntityList);
     } else if (isSMIRoute()) {
         routeBlueGreenSMI(kubectl, NONE_LABEL_VALUE, manifestObjects.deploymentEntityList, manifestObjects.serviceEntityList);
-        deleteWorkloadsWithLabel(kubectl, BLUE_GREEN_NEW_LABEL_VALUE, manifestObjects.deploymentEntityList);
+        deleteWorkloadsWithLabel(kubectl, GREEN_LABEL_VALUE, manifestObjects.deploymentEntityList);
         cleanupSMI(kubectl, manifestObjects.deploymentEntityList, manifestObjects.serviceEntityList);    
     } else {
         routeBlueGreenService(kubectl, NONE_LABEL_VALUE, manifestObjects.deploymentEntityList, manifestObjects.serviceEntityList);
-        deleteWorkloadsWithLabel(kubectl, BLUE_GREEN_NEW_LABEL_VALUE, manifestObjects.deploymentEntityList);
+        deleteWorkloadsWithLabel(kubectl, GREEN_LABEL_VALUE, manifestObjects.deploymentEntityList);
     }
 }
--- a/src/actions/promote.ts
+++ b/src/actions/promote.ts
@@ -10,10 +10,10 @@ import * as KubernetesObjectUtility from '../utilities/resource-object-utility';
 import * as models from '../constants';
 import * as KubernetesManifestUtility from '../utilities/manifest-stability-utility';
 import { getManifestObjects, deleteWorkloadsWithLabel, deleteWorkloadsAndServicesWithLabel } from '../utilities/strategy-helpers/blue-green-helper';
-import { BLUE_GREEN_NEW_LABEL_VALUE, NONE_LABEL_VALUE } from '../utilities/strategy-helpers/blue-green-helper';
-import { isBlueGreenDeploymentStrategy, blueGreenRouteService, blueGreenPromote } from '../utilities/strategy-helpers/service-blue-green-helper';
-import { isIngressRoute, blueGreenRouteIngress, blueGreenPromoteIngress } from '../utilities/strategy-helpers/ingress-blue-green-helper';
-import { isSMIRoute, blueGreenRouteTraffic, blueGreenPromoteSMI, cleanSetUpSMI } from '../utilities/strategy-helpers/smi-blue-green-helper';
+import { isBlueGreenDeploymentStrategy, isIngressRoute, isSMIRoute, BLUE_GREEN_NEW_LABEL_VALUE, NONE_LABEL_VALUE } from '../utilities/strategy-helpers/blue-green-helper';
+import { routeBlueGreenService, blueGreenPromote } from '../utilities/strategy-helpers/service-blue-green-helper';
+import { routeBlueGreenIngress, blueGreenPromoteIngress } from '../utilities/strategy-helpers/ingress-blue-green-helper';
+import { routeBlueGreenSMI, blueGreenPromoteSMI, cleanupSMI } from '../utilities/strategy-helpers/smi-blue-green-helper';
 import { Kubectl, Resource } from '../kubectl-object-model';
 
 export async function promote() {
@@ -79,14 +79,14 @@ async function promoteBlueGreen(kubectl: Kubectl) {
     core.debug('routing to new deployments');
     await KubernetesManifestUtility.checkManifestStability(kubectl, resourceTypes);
     if(isIngressRoute()) {
-        blueGreenRouteIngress(kubectl, null, manifestObjects.serviceNameMap, manifestObjects.serviceEntityList, manifestObjects.ingressEntityList);
+        routeBlueGreenIngress(kubectl, null, manifestObjects.serviceNameMap, manifestObjects.serviceEntityList, manifestObjects.ingressEntityList);
         deleteWorkloadsAndServicesWithLabel(kubectl, BLUE_GREEN_NEW_LABEL_VALUE, manifestObjects.deploymentEntityList, manifestObjects.serviceEntityList);
     } else if (isSMIRoute()) {
-        blueGreenRouteTraffic(kubectl, NONE_LABEL_VALUE, manifestObjects.deploymentEntityList, manifestObjects.serviceEntityList);
+        routeBlueGreenSMI(kubectl, NONE_LABEL_VALUE, manifestObjects.deploymentEntityList, manifestObjects.serviceEntityList);
         deleteWorkloadsWithLabel(kubectl, BLUE_GREEN_NEW_LABEL_VALUE, manifestObjects.deploymentEntityList);
-        cleanSetUpSMI(kubectl, manifestObjects.deploymentEntityList, manifestObjects.serviceEntityList);    
+        cleanupSMI(kubectl, manifestObjects.deploymentEntityList, manifestObjects.serviceEntityList);    
     } else {
-        blueGreenRouteService(kubectl, NONE_LABEL_VALUE, manifestObjects.deploymentEntityList, manifestObjects.serviceEntityList);
+        routeBlueGreenService(kubectl, NONE_LABEL_VALUE, manifestObjects.deploymentEntityList, manifestObjects.serviceEntityList);
         deleteWorkloadsWithLabel(kubectl, BLUE_GREEN_NEW_LABEL_VALUE, manifestObjects.deploymentEntityList);
     }
 }
--- a/src/actions/reject.ts
+++ b/src/actions/reject.ts
@@ -5,9 +5,10 @@ import * as SMICanaryDeploymentHelper from '../utilities/strategy-helpers/smi-ca
 import { Kubectl } from '../kubectl-object-model';
 import * as utils from '../utilities/manifest-utilities';
 import * as TaskInputParameters from '../input-parameters';
-import { isBlueGreenDeploymentStrategy, blueGreenReject } from '../utilities/strategy-helpers/service-blue-green-helper';
-import { isIngressRoute ,blueGreenRejectIngress } from '../utilities/strategy-helpers/ingress-blue-green-helper';
-import { isSMIRoute, blueGreenRejectSMI} from '../utilities/strategy-helpers/smi-blue-green-helper'
+import { blueGreenReject } from '../utilities/strategy-helpers/service-blue-green-helper';
+import { blueGreenRejectIngress } from '../utilities/strategy-helpers/ingress-blue-green-helper';
+import { blueGreenRejectSMI} from '../utilities/strategy-helpers/smi-blue-green-helper'
+import { isSMIRoute, isIngressRoute, isBlueGreenDeploymentStrategy } from '../utilities/strategy-helpers/blue-green-helper'
 import { getManifestFiles } from '../utilities/strategy-helpers/deployment-helper'
 
 export async function reject() {
--- a/src/actions/reject.ts
+++ b/src/actions/reject.ts
@@ -5,9 +5,9 @@ import * as SMICanaryDeploymentHelper from '../utilities/strategy-helpers/smi-ca
 import { Kubectl } from '../kubectl-object-model';
 import * as utils from '../utilities/manifest-utilities';
 import * as TaskInputParameters from '../input-parameters';
-import { blueGreenReject } from '../utilities/strategy-helpers/service-blue-green-helper';
-import { blueGreenRejectIngress } from '../utilities/strategy-helpers/ingress-blue-green-helper';
-import { blueGreenRejectSMI} from '../utilities/strategy-helpers/smi-blue-green-helper'
+import { rejectBlueGreenService } from '../utilities/strategy-helpers/service-blue-green-helper';
+import { rejectBlueGreenIngress } from '../utilities/strategy-helpers/ingress-blue-green-helper';
+import { rejectBlueGreenSMI } from '../utilities/strategy-helpers/smi-blue-green-helper'
 import { isSMIRoute, isIngressRoute, isBlueGreenDeploymentStrategy } from '../utilities/strategy-helpers/blue-green-helper'
 import { getManifestFiles } from '../utilities/strategy-helpers/deployment-helper'
 
@@ -39,10 +39,10 @@ async function rejectCanary(kubectl: Kubectl) {
 async function rejectBlueGreen(kubectl: Kubectl) {
     let inputManifestFiles: string[] = getManifestFiles(TaskInputParameters.manifests);
     if(isIngressRoute()) {
-        await blueGreenRejectIngress(kubectl, inputManifestFiles);
+        await rejectBlueGreenIngress(kubectl, inputManifestFiles);
     } else if (isSMIRoute()) {
-        await blueGreenRejectSMI(kubectl, inputManifestFiles);
+        await rejectBlueGreenSMI(kubectl, inputManifestFiles);
     } else {
-        await blueGreenReject(kubectl, inputManifestFiles);
+        await rejectBlueGreenService(kubectl, inputManifestFiles);
     }
 }
--- a/src/actions/reject.ts
+++ b/src/actions/reject.ts
@@ -5,15 +5,25 @@ import * as SMICanaryDeploymentHelper from '../utilities/strategy-helpers/smi-ca
 import { Kubectl } from '../kubectl-object-model';
 import * as utils from '../utilities/manifest-utilities';
 import * as TaskInputParameters from '../input-parameters';
+import { isBlueGreenDeploymentStrategy, blueGreenReject } from '../utilities/strategy-helpers/service-blue-green-helper';
+import { isIngressRoute ,blueGreenRejectIngress } from '../utilities/strategy-helpers/ingress-blue-green-helper';
+import { isSMIRoute, blueGreenRejectSMI} from '../utilities/strategy-helpers/smi-blue-green-helper'
+import { getManifestFiles } from '../utilities/strategy-helpers/deployment-helper'
 
-export async function reject(ignoreSslErrors?: boolean) {
-    const kubectl = new Kubectl(await utils.getKubectl(), TaskInputParameters.namespace, ignoreSslErrors);
+export async function reject() {
+    const kubectl = new Kubectl(await utils.getKubectl(), TaskInputParameters.namespace, true);
 
-    if (!canaryDeploymentHelper.isCanaryDeploymentStrategy()) {
-        core.debug('Strategy is not canary deployment. Invalid request.');
-        throw ('InvalidRejectActionDeploymentStrategy');
+    if (canaryDeploymentHelper.isCanaryDeploymentStrategy()) {
+        await rejectCanary(kubectl);
+    } else if (isBlueGreenDeploymentStrategy()) {
+        await rejectBlueGreen(kubectl);
+    } else {
+        core.debug('Strategy is not canary or blue-green deployment. Invalid request.');
+        throw ('InvalidDeletetActionDeploymentStrategy');
     }
+}
 
+async function rejectCanary(kubectl: Kubectl) {
     let includeServices = false;
     if (canaryDeploymentHelper.isSMICanaryStrategy()) {
         core.debug('Reject deployment with SMI canary strategy');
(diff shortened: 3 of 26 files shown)


ds-ms:
Add a proper description

action.yml
```diff
@@ -19,9 +19,17 @@ inputs:
...
   strategy:
-    description: 'Deployment strategy to be used. Allowed values are none, canary'
+    description: 'Deployment strategy to be used. Allowed values are none, canary and blue-green'
     required: false
     default: 'none'
+  route-method:
+    description: 'Route based on service, ingress or SMI for blue-green strategy'
+    required: false
+    default: 'service'
+  version-switch-buffer:
```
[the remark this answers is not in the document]
ds-ms: What does buffer time mean here?

src/run.ts
```diff
@@ -67,10 +67,20 @@ export async function run() {
         await deploy(new Kubectl(kubectlPath, namespace), manifests, strategy);
     }
     else if (action === 'promote') {
-        await promote(true);
+        let strategy = core.getInput('strategy');
+        if(strategy==='canary') {
```
ds-ms, line 71: Fix formatting

src/utilities/resource-object-utility.ts
```diff
@@ -34,6 +34,14 @@ export function isServiceEntity(kind: string): boolean {
     return isEqual("Service", kind, StringComparer.OrdinalIgnoreCase);
 }
 
+export function isIngressEntity(kind: string): boolean {
+    if(!kind) {
```
ds-ms, line 38: NIT: Formatting

src/utilities/strategy-helpers/deployment-helper.ts
```diff
@@ -16,7 +16,11 @@ import { Kubectl, Resource } from '../../kubectl-object-model';
 
 import { deployPodCanary } from './pod-canary-deployment-helper';
 import { deploySMICanary } from './smi-canary-deployment-helper';
-import { checkForErrors } from "../utility";
+import { checkForErrors, sleep } from "../utility";
+
+//bg
```
ds-ms, line 21: Don't use abbreviations

src/utilities/strategy-helpers/deployment-helper.ts
```diff
@@ -65,6 +101,17 @@ function deployManifests(files: string[], kubectl: Kubectl, isCanaryDeploymentSt
         }
         result = canaryDeploymentOutput.result;
         files = canaryDeploymentOutput.newFilePaths;
+    } else if(isBlueGreenDeploymentStrategy) {
```
ds-ms, line 104: Run a formatter against all the files

src/utilities/strategy-helpers/deployment-helper.ts
```diff
@@ -31,19 +35,51 @@ export async function deploy(kubectl: Kubectl, manifestFilePaths: string[], depl
...
+    let sleepTime: number = TaskInputParameters.versionSwitchBuffer;
+    // check if proper input
+    if(sleepTime<0 || sleepTime>300) {
+        console.log('Max buffer time is 300 minutes');
+        throw('TooLongBufferTime')
+    }
+
+    let temp = new Date();
+    console.log('starting buffer time of '+sleepTime+' minute/s at '+temp.getHours()+':'+temp.getMinutes()+':'+temp.getSeconds()+' UTC');
+    // waiting
```
ds-ms, line 67: Move the console log statements to `sleep` function.
Where you log the start time and end time

src/utilities/strategy-helpers/deployment-helper.ts
```diff
@@ -31,19 +35,51 @@ export async function deploy(kubectl: Kubectl, manifestFilePaths: string[], depl
...
+    await sleep(sleepTime*1000*60);
+    temp = new Date();
+    console.log('stopping buffer time of '+sleepTime+' minute/s at '+temp.getHours()+':'+temp.getMinutes()+':'+temp.getSeconds()+' UTC');
+    
+    // routing to new deployments
+    if(blueGreenDeploymentHelper.isIngressRoute()) {
+        blueGreenDeploymentHelper.blueGreenRouteIngress(kubectl, inputManifestFiles, blueGreenDeploymentHelper.BLUE_GREEN_NEW_LABEL_VALUE);
```
ds-ms, line 74: `VALUE` should be implicit right? Just use the word `BLUE_GREEN_NEW_LABEL`

src/utilities/strategy-helpers/deployment-helper.ts
```diff
@@ -31,19 +35,51 @@ export async function deploy(kubectl: Kubectl, manifestFilePaths: string[], depl
...
     ingressResources.forEach(ingressResource => {
         kubectl.getResource(KubernetesConstants.DiscoveryAndLoadBalancerResource.ingress, ingressResource.name);
     });
 }
 
+async function routeBlueGreen(kubectl: Kubectl, inputManifestFiles: string[]) {
+    // get buffer time
+    let sleepTime: number = TaskInputParameters.versionSwitchBuffer;
+    // check if proper input
+    if(sleepTime<0 || sleepTime>300) {
```
ds-ms, line 60: `sleepTime = 0` is not allowed?

src/utilities/strategy-helpers/deployment-helper.ts
```diff
@@ -31,19 +35,51 @@ export async function deploy(kubectl: Kubectl, manifestFilePaths: string[], depl
...
+        console.log('Max buffer time is 300 minutes');
+        throw('TooLongBufferTime')
```
ds-ms, line 62: `TooLongBufferTime` doesn't hold true for `<0` cases. Keep them separate

src/utilities/strategy-helpers/deployment-helper.ts
```diff
@@ -31,19 +35,51 @@ export async function deploy(kubectl: Kubectl, manifestFilePaths: string[], depl
...
+    const deployedManifestFiles = deployManifests(inputManifestFiles, kubectl, isCanaryDeploymentStrategy(deploymentStrategy), isBlueGreenDeploymentStrategy(deploymentStrategy));
 
     // check manifest stability
     const resourceTypes: Resource[] = KubernetesObjectUtility.getResources(deployedManifestFiles, models.deploymentTypes.concat([KubernetesConstants.DiscoveryAndLoadBalancerResource.service]));
     await checkManifestStability(kubectl, resourceTypes);
 
+    // route blue-green deployments
+    if (blueGreenDeploymentHelper.isBlueGreenDeploymentStrategy()) {
+        await routeBlueGreen(kubectl, inputManifestFiles);
+    }
```
ds-ms, line 47: Shouldn't this logic be in `deployManifests` function?

src/run.ts
```diff
@@ -67,10 +67,20 @@ export async function run() {
         await deploy(new Kubectl(kubectlPath, namespace), manifests, strategy);
     }
     else if (action === 'promote') {
-        await promote(true);
+        let strategy = core.getInput('strategy');
```
ds-ms, line 70: Convert it `toLowerCase` or use the `isEqual` function in the expressions below

src/run.ts
```diff
@@ -67,10 +67,20 @@ export async function run() {
...
+        if(strategy==='canary') {
+            await promote(true);
+        } else {
+            await promoteBlueGreen(true);
+        }
     }
     else if (action === 'reject') {
-        await reject(true);
+        let strategy = core.getInput('strategy');
```
ds-ms, line 78: Get the strategy value in a higher scope.. to avoid having multiple places with the same logic

src/run.ts
```diff
@@ -67,10 +67,20 @@ export async function run() {
         await deploy(new Kubectl(kubectlPath, namespace), manifests, strategy);
     }
     else if (action === 'promote') {
-        await promote(true);
+        let strategy = core.getInput('strategy');
+        if(strategy==='canary') {
+            await promote(true);
```
ds-ms, line 72: I don't see you sending any other input.. if it's true by default; why would you want to have a parameter?

ds-ms: Same applies for all the newly introduced functions

src/actions/reject.ts
```diff
@@ -23,4 +24,34 @@ export async function reject(ignoreSslErrors?: boolean) {
...
+    } else {
+        core.debug('deleting new deployment');
+        await blueGreenDeploymentHelper.blueGreenReject(kubectl, inputManifestFiles);
+    }
+}
+
+function getManifestFiles(manifestFilePaths: string[]): string[] {
+    const files: string[] = utils.getManifestFiles(manifestFilePaths);
+
+    if (files == null || files.length === 0) {
```
ds-ms, line 52: `if (!files)`

ds-ms: image

src/actions/reject.ts
```diff
@@ -23,4 +24,34 @@ export async function reject(ignoreSslErrors?: boolean) {
 
     core.debug('Deployment strategy selected is Canary. Deleting baseline and canary workloads.');
     canaryDeploymentHelper.deleteCanaryDeployment(kubectl, TaskInputParameters.manifests, includeServices);
+}
+
+export async function rejectBlueGreen(ignoreSslErrors?: boolean) {
```
ds-ms, line 29: Why'd you choose to have a separate function here?
You could have added the switch on values in the `reject` function right? That way there'd be a single entry point. 
You just call `reject` from the run function.. similar to `deploy` function and it takes care of the strategy and stuff internally.

ds-ms: Same would apply for `promote`

src/actions/reject.ts
ds-ms: Also, what's the point of this wrapper function? You could have added the throw logic in the utils function only right

src/actions/promote.ts
```diff
@@ -41,4 +50,114 @@ export async function promote(ignoreSslErrors?: boolean) {
...
+        core.debug('Strategy is not blue-green deployment. Invalid request.');
+        throw ('InvalidPromotetActionDeploymentStrategy');
+    }
+
+    let inputManifestFiles: string[] = getManifestFiles(TaskInputParameters.manifests);
+
+    inputManifestFiles = updateContainerImagesInManifestFiles(inputManifestFiles, TaskInputParameters.containers);
+
+    // imagePullSecrets addition
+    inputManifestFiles = updateImagePullSecretsInManifestFiles(inputManifestFiles, TaskInputParameters.imagePullSecrets);
```
ds-ms, line 68: This is duplicate code across the entire action; consider moving it to a util file

src/actions/promote.ts
```diff
@@ -41,4 +50,114 @@ export async function promote(ignoreSslErrors?: boolean) {
...
+
+    // checking if rollouts finished and then routing
+    if(blueGreenDeploymentHelper.isIngressRoute()) {
+        let result = await blueGreenDeploymentHelper.blueGreenPromoteIngress(kubectl, inputManifestFiles);
+        const deployedManifestFiles = result.newFilePaths;
+        const resourceTypes: Resource[] = KubernetesObjectUtility.getResources(deployedManifestFiles, models.deploymentTypes.concat([models.DiscoveryAndLoadBalancerResource.service]));
+        await checkManifestStability(kubectl, resourceTypes);
```
ds-ms, line 75: Duplicate code.. move this to a common function

src/actions/promote.ts
```diff
@@ -41,4 +50,114 @@ export async function promote(ignoreSslErrors?: boolean) {
...
+        blueGreenDeploymentHelper.cleanSetUpSMI(kubectl, inputManifestFiles);    
+    } else {
+        core.debug('deleting old deployment and making new ones');
+        let result = await blueGreenDeploymentHelper.blueGreenPromote(kubectl, inputManifestFiles);
+        const deployedManifestFiles = result.newFilePaths;
+        const resourceTypes: Resource[] = KubernetesObjectUtility.getResources(deployedManifestFiles, models.deploymentTypes.concat([models.DiscoveryAndLoadBalancerResource.service]));
+        await checkManifestStability(kubectl, resourceTypes);
+        core.debug('routing to new deployments');
+        blueGreenDeploymentHelper.blueGreenRouteService(kubectl, inputManifestFiles, blueGreenDeploymentHelper.NONE_LABEL_VALUE);
+        blueGreenDeploymentHelper.deleteOpposite(kubectl, inputManifestFiles, blueGreenDeploymentHelper.BLUE_GREEN_NEW_LABEL_VALUE);
```
ds-ms, line 94: Consider using a different function name.. `deleteOpposite` doesn't set a proper context

src/actions/promote.ts
```diff
@@ -41,4 +50,114 @@ export async function promote(ignoreSslErrors?: boolean) {
...
+    }
+}
+
+async function checkManifestStability(kubectl: Kubectl, resources: Resource[]): Promise<void> {
+    await KubernetesManifestUtility.checkManifestStability(kubectl, resources);
+}
```
ds-ms, line 100: Why? You could have directly used `await KubernetesManifestUtility.checkManifestStability(kubectl, resources)` at the invoker

src/actions/promote.ts
```diff
@@ -41,4 +50,114 @@ export async function promote(ignoreSslErrors?: boolean) {
...
+
+function getManifestFiles(manifestFilePaths: string[]): string[] {
+    const files: string[] = utils.getManifestFiles(manifestFilePaths);
+
+    if (files == null || files.length === 0) {
+        throw new Error(`ManifestFileNotFound : ${manifestFilePaths}`);
+    }
+
+    return files;
+}
```
ds-ms, line 111: Remove this function too.. I see you've written a lot of wrapper functions

src/actions/promote.ts
```diff
@@ -41,4 +50,114 @@ export async function promote(ignoreSslErrors?: boolean) {
...
+            );
+            newFilePaths.push(fileName);
+        });
+
+        return newFilePaths;
+    }
+
+    return filePaths;
+}
+
```
ds-ms, line 142: Does this do anything different from the common one? 
You could have called them using different inputs right?

src/actions/promote.ts
```diff
@@ -41,4 +50,114 @@ export async function promote(ignoreSslErrors?: boolean) {
...
+                    }
+                    newObjectsList.push(inputObject);
+                }
+            });
+        });
+        core.debug('New K8s objects after addin imagePullSecrets are :' + JSON.stringify(newObjectsList));
+        const newFilePaths = fileHelper.writeObjectsToFile(newObjectsList);
+        return newFilePaths;
+    }
+    return filePaths;
```
ds-ms, line 162: This is duplicate code. You could use the same logic from the utils class

action.yml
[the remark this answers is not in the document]
sundargs2000: no this is the time where a deployment is made but live traffic isn't routed to it. the user can test the created deployments during this time.

src/utilities/strategy-helpers/deployment-helper.ts
sundargs2000: it is allowed. thats the default value.

src/run.ts
sundargs2000: promote already exists right? even i was confused about the parameter. ill change and remove it

src/utilities/strategy-helpers/deployment-helper.ts
sundargs2000: im using an already existing sleep function from utils which is used in manifest stability file where i dont think logging is a good idea. should i create another function?

src/utilities/strategy-helpers/deployment-helper.ts
sundargs2000: im routing only after deployments have rolled out successfully

src/actions/promote.ts
sundargs2000: im deleting the deployment im not targeting now, ive used it in `blue-green-helper` to delete stable versions. should i change it to deleteColor or any other suggestion?

sundargs2000:
Hey guys, based on the comments I got so far, I have changed some things, (the implementation logic is same, I have reduced redundancy and optimised , and better test descriptions).  Let me know if I should squash these commits into one.

ds-ms:
@sundargs2000 just squash them before merging.. 😃

sundargs2000:
ok done

src/utilities/strategy-helpers/blue-green-helper.ts
```diff
@@ -0,0 +1,822 @@
...
+    return deploymentStrategy && deploymentStrategy.toUpperCase() === BLUE_GREEN_DEPLOYMENT_STRATEGY;
+}
+
+export function deployBlueGreen(kubectl: Kubectl, filePaths: string[]) {
+
+    // get all kubernetes objects defined in manifest files
+    const manifestObjects = getManifestObjects(filePaths);
+
+    // checking if there are existing green deployments with green label value
+    shouldWeDeploy(kubectl, manifestObjects.depEntList);
```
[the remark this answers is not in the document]
ds-ms: Also, name the function something like.. `checkCurrentDeploymentState` or `isGreenDeployed` or something

src/utilities/strategy-helpers/blue-green-helper.ts
```diff
@@ -0,0 +1,822 @@
...
+    const manifestObjects = getManifestObjects(filePaths);
+
+    // routing to stable objects
+    blueGreenRouteService(kubectl, NONE_LABEL_VALUE, manifestObjects.depEntList, manifestObjects.serEntList);
+
+    // seeing if we should even delete the service
+    cleanUp(kubectl, manifestObjects.depEntList, manifestObjects.serEntList);
+    
+    // deleting the new deployments with green suffix
+    deleteOpposite(kubectl, BLUE_GREEN_NEW_LABEL_VALUE, manifestObjects.depEntList);
```
ds-ms, line 89: Can this deleteOpposite's logic be a part of the `cleanUp` function itself?

src/utilities/strategy-helpers/blue-green-helper.ts
```diff
@@ -0,0 +1,822 @@
...
+export async function blueGreenPromote(kubectl: Kubectl, filePaths: string[]) {
+
+    // get all kubernetes objects defined in manifest files
+    const manifestObjects = getManifestObjects(filePaths);
+
+    // checking if services are in the right state ie. targeting green deployments
+    shouldWePromote(kubectl, manifestObjects.depEntList, manifestObjects.serEntList);
+
+    // deleting previous stable deployments
+    deleteOpposite(kubectl, NONE_LABEL_VALUE, manifestObjects.depEntList);
```
ds-ms, line 68: I'm not particularly fond of this function name `deleteOpposite` 😛  It doesn't seem to improve readability?

src/utilities/strategy-helpers/blue-green-helper.ts
```diff
@@ -0,0 +1,822 @@
...
+
+export function blueGreenRouteService(kubectl: Kubectl, nextLabel: string, depEntList: any[], serEntList: any[]) {
+    
+    const newObjectsList = [];
+    serEntList.forEach((inputObject) => {
+        let isRouted: boolean = false;
+        depEntList.forEach((depObject) => {
+            // finding if there is a deployment in the given manifests the service targets
+            if (getServiceSelector(inputObject) && getDeploymentMatchLabels(depObject) && getServiceSelector(inputObject) === getDeploymentMatchLabels(depObject)) {
+                isRouted = true;
```
ds-ms, line 117: Don't call these functions inline

src/utilities/strategy-helpers/blue-green-helper.ts
```diff
@@ -0,0 +1,822 @@
...
+        try {
+            const result = kubectl.delete([delObject.kind, delObject.name]);
+            checkForErrors([result]);
+        } catch (ex) {
+            // Ignore failures of delete if doesn't exist
+        }
+    });
+}
+
+export function cleanUp(kubectl: Kubectl, depEntList: any[], serEntList: any[]) {
```
ds-ms, line 193: Use full names here as well

ds-ms:
This is a really big PR.  I strongly suggest getting at least one more pair of  👀 on this PR so that we don't miss out on some key things

cc: @ajinkya599 / @shigupt202 could you review this once

src/actions/promote.ts
```diff
@@ -7,38 +7,78 @@ import * as SMICanaryDeploymentHelper from '../utilities/strategy-helpers/smi-ca
...
+        const deployedManifestFiles = result.newFilePaths;
+        const resourceTypes: Resource[] = KubernetesObjectUtility.getResources(deployedManifestFiles, models.deploymentTypes.concat([models.DiscoveryAndLoadBalancerResource.service]));
+        await KubernetesManifestUtility.checkManifestStability(kubectl, resourceTypes);
+        
+        const manifestObjects = blueGreenDeploymentHelper.getManifestObjects(inputManifestFiles);
+        core.debug('routing to new deployments');
+        await KubernetesManifestUtility.checkManifestStability(kubectl, resourceTypes);
+        if(blueGreenDeploymentHelper.isIngressRoute()) {
+            blueGreenDeploymentHelper.blueGreenRouteIngress(kubectl, null, manifestObjects.serEntSet, manifestObjects.serEntList, manifestObjects.ingEntList);
+            blueGreenDeploymentHelper.deleteOppositeIngress(kubectl, blueGreenDeploymentHelper.BLUE_GREEN_NEW_LABEL_VALUE, manifestObjects.depEntList, manifestObjects.serEntList);
```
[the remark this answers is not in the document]
sundargs2000: cleanup sees whether a service needs to be deleted after a rejection, deletopposite just deletes deployments of given label

sundargs2000:
yes lets have a call. ill ask ajinkya to join too.

src/utilities/strategy-helpers/blue-green-helper.ts
sundargs2000, line 45: remove

src/utilities/strategy-helpers/blue-green-helper.ts
```diff
@@ -0,0 +1,822 @@
...
+export function deployBlueGreen(kubectl: Kubectl, filePaths: string[]) {
+
+    // get all kubernetes objects defined in manifest files
+    const manifestObjects = getManifestObjects(filePaths);
+
+    // checking if there are existing green deployments with green label value
+    shouldWeDeploy(kubectl, manifestObjects.depEntList);
+
+    // create deployments with green label value
+    const result = createDeploymentsWithLabel(kubectl, manifestObjects.depEntList, BLUE_GREEN_NEW_LABEL_VALUE);
```
sundargs2000, line 48: change to workloads

src/utilities/strategy-helpers/blue-green-helper.ts
shigupt202: Agreed. `deleteStableWorkloads` or `deleteCurrentStableWorkloads` maybe?

src/utilities/strategy-helpers/blue-green-helper.ts
```diff
@@ -0,0 +1,822 @@
...
+    return result;
+}
+
+export async function blueGreenPromote(kubectl: Kubectl, filePaths: string[]) {
+
+    // get all kubernetes objects defined in manifest files
+    const manifestObjects = getManifestObjects(filePaths);
+
+    // checking if services are in the right state ie. targeting green deployments
+    shouldWePromote(kubectl, manifestObjects.depEntList, manifestObjects.serEntList);
```
shigupt202, line 65: Change it to `validateServiceState` or you can probably make it a boolean function `isServiceTargettingWorkloads` or something like that. You'll have to add an if-else block here in that case.

src/utilities/strategy-helpers/blue-green-helper.ts
```diff
@@ -0,0 +1,822 @@
...
+    blueGreenRouteService(kubectl, NONE_LABEL_VALUE, manifestObjects.depEntList, manifestObjects.serEntList);
+
+    // seeing if we should even delete the service
+    cleanUp(kubectl, manifestObjects.depEntList, manifestObjects.serEntList);
+    
+    // deleting the new deployments with green suffix
+    deleteOpposite(kubectl, BLUE_GREEN_NEW_LABEL_VALUE, manifestObjects.depEntList);
+}
+
+export function shouldWeDeploy(kubectl: Kubectl, depEntList: any[]) {
```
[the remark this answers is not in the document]
shigupt202: Also, you can change the function name to `stableBlueGreenWorkloadExists` and make it boolean.

src/actions/promote.ts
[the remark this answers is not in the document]
shigupt202: How about we call it `deleteService` but add the required condition before calling it?

shigupt202:
Make sure to check existing flows, like canary with all the strategies and actions, because there have been changes in already existing code.

src/run.ts
```diff
@@ -67,10 +67,10 @@ export async function run() {
         await deploy(new Kubectl(kubectlPath, namespace), manifests, strategy);
     }
     else if (action === 'promote') {
-        await promote(true);
+        await promote();
```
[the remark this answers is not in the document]
sundargs2000: promote was only called from so the param was always true

src/utilities/strategy-helpers/ingress-blue-green-helper.ts
```diff
@@ -0,0 +1,179 @@
...
+import { createWorkloadssWithLabel, getManifestObjects, getNewBlueGreenObject, addBlueGreenLabelsAndAnnotations, deleteWorkloadsAndServicesWithLabel, fetchResource } from './blue-green-helper';
+import { BLUE_GREEN_NEW_LABEL_VALUE, NONE_LABEL_VALUE, BLUE_GREEN_VERSION_LABEL } from './blue-green-helper';
+
+const INGRESS_ROUTE = 'INGRESS';
+
+export function isIngressRoute(): boolean {
+    const routeMethod = TaskInputParameters.routeMethod;
+    return routeMethod && routeMethod.toUpperCase() === INGRESS_ROUTE;
+}
+export function deployBlueGreenIngress(kubectl: Kubectl, filePaths: string[]) {
```
ajinkya599, line 16: add newline

src/utilities/strategy-helpers/ingress-blue-green-helper.ts
```diff
@@ -0,0 +1,179 @@
...
+
+    // Updating ingress labels
+    let finalObject =  updateIngressBackend(newObject, serviceNameMap);
+    return finalObject;
+}
+
+export function updateIngressBackend(inputObject: any, serviceNameMap: Map<string, string>): any {
+    inputObject = JSON.parse(JSON.stringify(inputObject), (key, value) => {
+        if(key.toUpperCase() === 'BACKEND') {
+            let serName = value.serviceName;
```
ajinkya599, line 169: serviceName

src/utilities/strategy-helpers/ingress-blue-green-helper.ts
```diff
@@ -0,0 +1,179 @@
...
+    addBlueGreenLabelsAndAnnotations(newObject, type);
+
+    // Updating ingress labels
+    let finalObject =  updateIngressBackend(newObject, serviceNameMap);
+    return finalObject;
+}
+
+export function updateIngressBackend(inputObject: any, serviceNameMap: Map<string, string>): any {
+    inputObject = JSON.parse(JSON.stringify(inputObject), (key, value) => {
+        if(key.toUpperCase() === 'BACKEND') {
```
ajinkya599, line 168: Define 'BACKEND' as a constant

src/utilities/strategy-helpers/ingress-blue-green-helper.ts
```diff
@@ -0,0 +1,179 @@
...
+            let serName = value.serviceName; 
+            if (serviceNameMap.has(serName)) {
+                //updating srvice name with corresponging bluegreen name only if service is provied in given manifests
```
ajinkya599, line 171: Typos:
service
corresponding

src/utilities/strategy-helpers/smi-blue-green-helper.ts
```diff
@@ -0,0 +1,241 @@
...
+                    } 
+                });
+            }
+        });
+    });               
+    
+    return isTrafficSplitInRightState;
+}
+
+export function cleanSetUpSMI(kubectl: Kubectl, deploymentEntityList: any[], serviceEntityList: any[]) {
```
ajinkya599, line 218: cleanupSMI

src/utilities/strategy-helpers/smi-blue-green-helper.ts
```diff
@@ -0,0 +1,241 @@
...
+    
+    const delList = [];
```
ajinkya599, line 220: deleteList or deletionList

src/utilities/strategy-helpers/smi-blue-green-helper.ts
```diff
@@ -0,0 +1,241 @@
...
+    // creating services and other objects
+    const newObjectsList = manifestObjects.otherObjects.concat(manifestObjects.serviceEntityList).concat(manifestObjects.ingressEntityList);
+    const manifestFiles = fileHelper.writeObjectsToFile(newObjectsList);
+    kubectl.apply(manifestFiles);
+
+    // make extraservices and trafficsplit
+    setUpSMI(kubectl, manifestObjects.deploymentEntityList, manifestObjects.serviceEntityList);
+
+    // create new deloyments
+    const result = createWorkloadssWithLabel(kubectl, manifestObjects.deploymentEntityList, BLUE_GREEN_NEW_LABEL_VALUE);
```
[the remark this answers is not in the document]
sundargs2000: so basically i wont be changing the routed service in smi. so if i create a new deployment before setting up, the routed service would be targeting both old and new deployments. in service and ingress, i am leaving a none label in service and deployment.

src/utilities/strategy-helpers/smi-blue-green-helper.ts
```diff
@@ -0,0 +1,241 @@
...
+    const trafficObjectList = []
+    serviceEntityList.forEach((inputObject) => {
+        deploymentEntityList.forEach((depObject) => {
+            // finding out whether service targets a deployment in given manifests
+            if (getServiceSelector(inputObject) && getDeploymentMatchLabels(depObject) && getServiceSelector(inputObject) === getDeploymentMatchLabels(depObject)) {
+                // decided that this service needs to be routed
+                //querying for both services
+                trafficObjectList.push(inputObject);
+                // setting up the services for trafficsplit
+                const newStableService = getSMIServiceResource(inputObject, STABLE_SUFFIX, 0);
```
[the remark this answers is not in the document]
sundargs2000: as i've replied before, initially when we need to make the routed service target old deployment, we'll be needing stable service then

src/utilities/strategy-helpers/service-blue-green-helper.ts
```diff
@@ -0,0 +1,130 @@
...
+            // finding if there is a deployment in the given manifests the service targets
+            if (getServiceSelector(inputObject) && getDeploymentMatchLabels(depObject) && getServiceSelector(inputObject) === getDeploymentMatchLabels(depObject)) {
+                isRouted = true;
+                // decided that this service needs to be routed
+                // point to the given nextlabel
+                const newBlueGreenServiceObject = getUpdatedBlueGreenService(inputObject, nextLabel);
+                newObjectsList.push(newBlueGreenServiceObject);
+            }
+        });
+        if (!isRouted) {
```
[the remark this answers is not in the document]
sundargs2000: hmm. yes. that is what i am doing in the new change. ill have routedservices and unrouted services separately which ill get from 'getManifestObjects' itself

src/utilities/manifest-utilities.ts
```diff
@@ -192,21 +198,34 @@ function substituteImageNameInSpecContent(currentString: string, imageName: stri
     }, '');
 }
 
-export function updateContainerImagesInManifestFiles(contents, containers: string[]): string {
```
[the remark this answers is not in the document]
sundargs2000: nope. a function with same name was rewritten was written in deployment helper and only that was being referenced. i moved the thing from deployment helper and using it everywhere.

src/utilities/strategy-helpers/blue-green-helper.ts
```diff
@@ -0,0 +1,271 @@
...
+                    ingressEntityList.push(inputObject);
+                } else {
+                    otherEntitiesList.push(inputObject);
+                }
+            }
+        });
+    })
+
+    let serviceNameMap = new Map<string, string>();
+    // find all services and adding they names with blue green suffix
```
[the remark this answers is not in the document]
sundargs2000: i would have to use a set (felt that list includes might be costly) so thought i might use a map itself. are you suggesting use a list itself?
.

src/utilities/strategy-helpers/smi-blue-green-helper.ts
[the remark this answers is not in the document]
ajinkya599: Let's discuss this once.

src/utilities/strategy-helpers/blue-green-helper.ts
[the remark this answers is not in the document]
ajinkya599: I was suggesting to rephrase the comment.

src/utilities/strategy-helpers/smi-blue-green-helper.ts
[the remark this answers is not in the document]
ajinkya599: let's discuss this once.

src/utilities/strategy-helpers/service-blue-green-helper.ts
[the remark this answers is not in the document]
ajinkya599: No. Let's target checking this PR in and take logic enhancements in the next iteration.

src/utilities/strategy-helpers/blue-green-helper.ts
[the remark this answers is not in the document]
sundargs2000: oh ok

src/utilities/strategy-helpers/service-blue-green-helper.ts
```diff
@@ -0,0 +1,130 @@
...
+}
+
+export function blueGreenRouteService(kubectl: Kubectl, nextLabel: string, deploymentEntityList: any[], serviceEntityList: any[]) {
+    
+    const newObjectsList = [];
+    serviceEntityList.forEach((inputObject) => {
+        let isRouted: boolean = false;
+        deploymentEntityList.forEach((depObject) => {
+            // finding if there is a deployment in the given manifests the service targets
+            if (getServiceSelector(inputObject) && getDeploymentMatchLabels(depObject) && getServiceSelector(inputObject) === getDeploymentMatchLabels(depObject)) {
```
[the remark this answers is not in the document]
sundargs2000: i have given a basic implementation for this, any refinements, will do in next

src/utilities/strategy-helpers/blue-green-helper.ts
```diff
@@ -0,0 +1,328 @@
...
+            deleteList.push(tempObject);
+        } else {
+            // else delete green labels
+            const tempObject = { name : name+BLUE_GREEN_SUFFIX, kind : kind };
+            deleteList.push(tempObject);
+        }
+    });
+
+    // delete services and deployments
+    deleteList.forEach((delObject) => {
```
ajinkya599, line 142: This code seems repeated in multiple places. Create a method and move to a common utility.

src/utilities/strategy-helpers/blue-green-helper.ts
```diff
@@ -0,0 +1,328 @@
...
+    deletionEntitiesList.forEach((inputObject) => {
+        const name = inputObject.metadata.name;
+        const kind = inputObject.kind;
+        if (!deleteLabel) {
+            // if not dellabel, delete stable objects
+            const tempObject = { name : name, kind : kind};
+            deleteList.push(tempObject);
+        } else {
+            // else delete green labels
+            const tempObject = { name : name+BLUE_GREEN_SUFFIX, kind : kind };
```
ajinkya599, line 136: Use `getBlueGreenResourceName`

src/utilities/strategy-helpers/blue-green-helper.ts
```diff
@@ -0,0 +1,328 @@
...
+export function deleteWorkloadsAndServicesWithLabel(kubectl: Kubectl, deleteLabel: string, deploymentEntityList: any[], serviceEntityList: any[]) {
+    // need to delete services and deployments
+    const deletionEntitiesList = deploymentEntityList.concat(serviceEntityList);
+    let deleteList = []
+    deletionEntitiesList.forEach((inputObject) => {
+        const name = inputObject.metadata.name;
+        const kind = inputObject.kind;
+        if (!deleteLabel) {
+            // if not dellabel, delete stable objects
+            const tempObject = { name : name, kind : kind};
```
ajinkya599, line 132: Rename to `resourceToDelete`

src/utilities/strategy-helpers/blue-green-helper.ts
```diff
@@ -0,0 +1,328 @@
...
+}
+
+export function deleteWorkloadsAndServicesWithLabel(kubectl: Kubectl, deleteLabel: string, deploymentEntityList: any[], serviceEntityList: any[]) {
+    // need to delete services and deployments
+    const deletionEntitiesList = deploymentEntityList.concat(serviceEntityList);
+    let deleteList = []
+    deletionEntitiesList.forEach((inputObject) => {
+        const name = inputObject.metadata.name;
+        const kind = inputObject.kind;
+        if (!deleteLabel) {
```
ajinkya599, line 130: Make use of `deleteLabel` in this method consistent with `deleteWorkloadsWithLabel`

src/utilities/strategy-helpers/smi-blue-green-helper.ts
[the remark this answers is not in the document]
ajinkya599: Unresolving for tracking.

src/utilities/strategy-helpers/blue-green-helper.ts
```diff
@@ -0,0 +1,271 @@
...
+
+export function getSpecLabel(inputObject: any): string {
+    if(!!inputObject && inputObject.spec && inputObject.spec.selector && inputObject.spec.selector.matchLabels && inputObject.spec.selector.matchLabels[BLUE_GREEN_VERSION_LABEL]) {
+        return inputObject.spec.selector.matchLabels[BLUE_GREEN_VERSION_LABEL]; 
+    }
+    return '';
+}
+
+export function getDeploymentMatchLabels(inputObject) {
+    if (inputObject.kind.toUpperCase()=='POD' && !!inputObject && !!inputObject.metadata && !!inputObject.metadata.labels) {
```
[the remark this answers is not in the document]
ajinkya599: Need to uppercase the constant `Pod`. Remove duplicate null check for `deploymentObject`

src/utilities/strategy-helpers/blue-green-helper.ts
```diff
@@ -0,0 +1,328 @@
...
+
+export function getSpecLabel(inputObject: any): string {
+    if(!!inputObject && inputObject.spec && inputObject.spec.selector && inputObject.spec.selector.matchLabels && inputObject.spec.selector.matchLabels[BLUE_GREEN_VERSION_LABEL]) {
+        return inputObject.spec.selector.matchLabels[BLUE_GREEN_VERSION_LABEL]; 
+    }
+    return '';
+}
+
+export function getDeploymentMatchLabels(inputObject): string {
+    if (inputObject.kind.toUpperCase()==KubernetesWorkload.pod && !!inputObject && !!inputObject.metadata && !!inputObject.metadata.labels) {
```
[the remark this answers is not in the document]
ajinkya599: Remove the second null check for `deploymentObject` as it's redundant

src/utilities/strategy-helpers/blue-green-helper.ts
```diff
@@ -0,0 +1,328 @@
...
+        try {
+            const result = kubectl.delete([delObject.kind, delObject.name]);
+            checkForErrors([result]);
+        } catch (ex) {
+            // Ignore failures of delete if doesn't exist
+        }
+    });
+}
+
+export function cleanUp(kubectl: Kubectl, deploymentEntityList: any[], serviceEntityList: any[]) {
```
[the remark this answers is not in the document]
ajinkya599: This was to handle first deployment reject scenario.
Not required as discussed offline.

src/utilities/strategy-helpers/deployment-helper.ts
```diff
@@ -13,38 +11,65 @@ import * as utils from '../manifest-utilities';
...
 }
 
-function getManifestFiles(manifestFilePaths: string[]): string[] {
+async function routeBlueGreen(kubectl: Kubectl, inputManifestFiles: string[]) {
+    // get buffer time
+    let sleepTime: number = parseInt(TaskInputParameters.versionSwitchBuffer);
+
+    //logging start of buffer time
+    let temp = new Date();
+    console.log('starting buffer time of '+sleepTime+' minute/s at '+temp.getHours()+':'+temp.getMinutes()+':'+temp.getSeconds()+' UTC');
```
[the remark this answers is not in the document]
anraghun: `console.log('Starting buffer time of <buffer-time> minute(s) at <timestamp>');`

src/utilities/strategy-helpers/smi-blue-green-helper.ts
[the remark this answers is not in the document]
ajinkya599: This makes sense.
Another thing came up while discussing this. We should find a solution for the case when the current deployment is not having none label.

src/utilities/strategy-helpers/smi-blue-green-helper.ts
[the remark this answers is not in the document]
ajinkya599: Discussed. We are creating temporary stable and green services for routing. Trafficsplit and these services get cleaned up eventually.

sundargs2000:
finally 🎉🎉 thanks guys

src/utilities/strategy-helpers/blue-green-helper.ts
```diff
@@ -0,0 +1,329 @@
...
+    if (!helper.isServiceEntity(inputObject.kind)) {
+        helper.updateSpecLabels(inputObject, newLabels, false);
+    }
+}
+
+export function getBlueGreenResourceName(name: string, suffix: string) {
+    return `${name}${suffix}`;
+}
+
+export function getSpecLabel(inputObject: any): string {
```
ajinkya599, line 243: Is `getSpecLabel` getting used anywhere?
If not, please cleanup. Else, consider passing label value as a parameter or change the method name

src/utilities/strategy-helpers/blue-green-helper.ts
```diff
@@ -0,0 +1,329 @@
...
+    if(!!inputObject && inputObject.spec && inputObject.spec.selector && inputObject.spec.selector.matchLabels && inputObject.spec.selector.matchLabels[BLUE_GREEN_VERSION_LABEL]) {
+        return inputObject.spec.selector.matchLabels[BLUE_GREEN_VERSION_LABEL]; 
+    }
+    return '';
+}
+
+export function getDeploymentMatchLabels(deploymentObject: any): string {
```
ajinkya599, line 250: Instead of returning `string`, return `any`.
Return the `matchLabels` object itself. Return null in place of `''`. Use it as a dictionary in the subsequent code.

src/utilities/strategy-helpers/blue-green-helper.ts
```diff
@@ -0,0 +1,329 @@
...
+    if (!!deploymentObject && deploymentObject.kind.toUpperCase()==KubernetesWorkload.pod.toUpperCase() &&  !!deploymentObject.metadata && !!deploymentObject.metadata.labels) {
+        return JSON.stringify(deploymentObject.metadata.labels);
+    } else if (!!deploymentObject && deploymentObject.spec && deploymentObject.spec.selector && deploymentObject.spec.selector.matchLabels) {
+        return JSON.stringify(deploymentObject.spec.selector.matchLabels);
+    }
+    return '';
+}
+
+export function getServiceSelector(serviceObject: any): string {
```
ajinkya599, line 259: Instead of returning `string`, return `any`.
Return the `spec selector` object itself. Return null in place of `''`. Use it as a dictionary in the subsequent code.

src/utilities/strategy-helpers/blue-green-helper.ts
```diff
@@ -0,0 +1,329 @@
...
+    if (!!serviceObject && serviceObject.spec && serviceObject.spec.selector) {
+        return JSON.stringify(serviceObject.spec.selector);
+    } else return '';
+}
+
+export function isServiceSelectorSubsetOfMatchLabel(serviceSelector: string, matchLabels: string): boolean {
```
ajinkya599, line 265: Get dictionary objects (any) in place of string.

src/utilities/strategy-helpers/blue-green-helper.ts
```diff
@@ -0,0 +1,329 @@
...
+  
+    JSON.parse(serviceSelector, (key, value) => {
+      serviceSelectorMap.set(key, value);
+    });
+    JSON.parse(matchLabels, (key, value) => {
+      matchLabelsMap.set(key, value);
+    });
+  
+    let isMatch = true;
+    serviceSelectorMap.forEach((value, key) => {
```
ajinkya599, line 277: Instead use `every`?

sundargs2000: seems like `every `is not supported for maps

PR #49, opened 2020-06-25
Checks in annotateNamespace to not error during failed annotation

koushdey:


lib/utilities/utility.js
```diff
@@ -72,16 +72,22 @@ function annotateChildPods(kubectl, resourceType, resourceName, allPods) {
...
 exports.annotateChildPods = annotateChildPods;
 function annotateNamespace(kubectl, namespaceName) {
-    let annotate = true;
     const result = kubectl.getResource('namespace', namespaceName);
-    this.checkForErrors([result]);
-    const annotationsSet = JSON.parse(result.stdout).metadata.annotations;
-    if (!!annotationsSet && !!annotationsSet.runUri && annotationsSet.runUri.indexOf(process.env['GITHUB_REPOSITORY']) == -1) {
-        annotate = false;
-        core.debug(`Skipping 'annotate namespace' as namespace annotated by other workflow`);
+    if (result == null) {
```
ds-ms, line 76: `if ( !result)`

lib/utilities/utility.js
```diff
@@ -72,16 +72,22 @@ function annotateChildPods(kubectl, resourceType, resourceName, allPods) {
...
+        return { code: -1, stderr: 'Failed to get resource' };
     }
-    if (annotate) {
-        return kubectl.annotate('namespace', namespaceName, constants_1.workflowAnnotations, true);
+    else if (!!result && !!result.stderr) {
+        return result;
+    }
+    if (!!result && !!result.stdout) {
+        const annotationsSet = JSON.parse(result.stdout).metadata.annotations;
```
ds-ms, line 83: Add null checks

lib/utilities/utility.js
```diff
@@ -72,16 +72,22 @@ function annotateChildPods(kubectl, resourceType, resourceName, allPods) {
...
+        if (!!annotationsSet && !!annotationsSet.runUri && annotationsSet.runUri.indexOf(process.env['GITHUB_REPOSITORY']) == -1) {
```
ds-ms, line 84: Extract the repository value outside; don't keep it inline

lib/utilities/utility.js
```diff
@@ -72,16 +72,22 @@ function annotateChildPods(kubectl, resourceType, resourceName, allPods) {
 }
 exports.annotateChildPods = annotateChildPods;
 function annotateNamespace(kubectl, namespaceName) {
-    let annotate = true;
     const result = kubectl.getResource('namespace', namespaceName);
-    this.checkForErrors([result]);
-    const annotationsSet = JSON.parse(result.stdout).metadata.annotations;
-    if (!!annotationsSet && !!annotationsSet.runUri && annotationsSet.runUri.indexOf(process.env['GITHUB_REPOSITORY']) == -1) {
-        annotate = false;
-        core.debug(`Skipping 'annotate namespace' as namespace annotated by other workflow`);
```
ds-ms, line 81: Why are we skipping here btw?

What if there are 2 repos targeting the same namespace? Would we only show traceability for one pipeline?

src/utilities/utility.ts
```diff
@@ -74,16 +74,23 @@ export function annotateChildPods(kubectl: Kubectl, resourceType: string, resour
...
 
 export function annotateNamespace(kubectl: Kubectl, namespaceName: string): IExecSyncResult {
-    let annotate = true;
     const result = kubectl.getResource('namespace', namespaceName);
-    this.checkForErrors([result]);
-    const annotationsSet = JSON.parse(result.stdout).metadata.annotations;
-    if (!!annotationsSet && !!annotationsSet.runUri && annotationsSet.runUri.indexOf(process.env['GITHUB_REPOSITORY']) == -1) {
-        annotate = false;
-        core.debug(`Skipping 'annotate namespace' as namespace annotated by other workflow`);
+    if (result == null) {
```
ds-ms, line 78: `if (!result)`

src/utilities/utility.ts
```diff
@@ -74,16 +74,23 @@ export function annotateChildPods(kubectl: Kubectl, resourceType: string, resour
...
+        return <IExecSyncResult>{ code: -1, stderr: 'Failed to get resource' };
```
ds-ms, line 79: use `as IExecSyncResult` instead?

src/utilities/utility.ts
```diff
@@ -74,16 +74,23 @@ export function annotateChildPods(kubectl: Kubectl, resourceType: string, resour
...
     }
-    if (annotate) {
-        return kubectl.annotate('namespace', namespaceName, workflowAnnotations, true);
+    else if (!!result && !!result.stderr) {
```
ds-ms, line 81: `!!` is redundant here right?

src/utilities/utility.ts
```diff
@@ -74,16 +74,23 @@ export function annotateChildPods(kubectl: Kubectl, resourceType: string, resour
...
+        return result;
+    }
+
+    if (!!result && !!result.stdout) {
+        const annotationsSet = JSON.parse(result.stdout).metadata.annotations;
```
ds-ms, line 86: Add null checks

lib/utilities/utility.js
koushdey: Yes, as per mvp, for differentiating two workflows we were taking an assumption of two different repos. In case of two workflows we will skip annotate namespace when we find repo annotations don't match. But I am working on adding a new condition that will look for the combination of repo + username + workflow two differentiate the workflows and we will not be skipping annotation at this time but actually append it in existing annotations as a json array.

lib/utilities/utility.js
koushdey: null checks on result and stdout added. Metadata always exists in the response. For annotations checks are there in next  line.

lib/utilities/utility.js
koushdey: Done

src/utilities/utility.ts
koushdey: Done

lib/utilities/utility.js
koushdey: Done

src/utilities/utility.ts
koushdey: Done

src/utilities/utility.ts
koushdey: responded above for the same.

src/utilities/utility.ts
koushdey: I checked for `!!` in conditions for JS. It ensures the resulting type is a boolean. [Ref](https://stackoverflow.com/questions/7452720/what-does-the-double-exclamation-operator-mean)

ds-ms: It does.. I agree..  but it's also redundant.. adds to value in this case

src/utilities/utility.ts
ds-ms: I meant the null checks for `.metadata` and `.annotations`

src/utilities/utility.ts
ds-ms: This `!!` thing is inconsistent across the file. So I suggested to remove it

src/utilities/utility.ts
koushdey: `metadata` is mandatory object and `annotation` check is added in next line.

src/utilities/utility.ts
koushdey: Will change it.

PR #52, opened 2020-07-15
Introduce annotations relevant to resource view during deploy

koushdey:


src/utilities/utility.ts
```diff
@@ -47,6 +50,50 @@ export function checkForErrors(execResults, warnIfError?: boolean) {
...
+        });
+    }
+
+    return commandExecutionResults;
+}
+
+export function annotateNamespace(kubectl: Kubectl, namespaceName: string): IExecSyncResult {
+    const result = kubectl.getResource('namespace', namespaceName);
+    if (!result) {
+        return { code: -1, stderr: 'Failed to get resource' } as IExecSyncResult;
```
ds-ms, line 79: Shouldn't the code be a non zero exit code and not a negative one?

ds-ms: make it 1

src/utilities/utility.ts
```diff
@@ -47,6 +50,50 @@ export function checkForErrors(execResults, warnIfError?: boolean) {
...
+    }
+    else if (result && result.stderr) {
+        return result;
+    }
+
+    if (result && result.stdout) {
```
ds-ms, line 85: This entire thing should go under the else block to avoid validating `if (result)` multiple times

src/utilities/utility.ts
```diff
@@ -47,6 +50,50 @@ export function checkForErrors(execResults, warnIfError?: boolean) {
...
+        owner = kubectl.getNewReplicaSet(resourceName);
+    }
+
+    if (allPods && allPods.items && allPods.items.length > 0) {
+        allPods.items.forEach((pod) => {
+            const owners = pod.metadata.ownerReferences;
+            if (owners) {
+                owners.forEach(ownerRef => {
+                    if (ownerRef.name === owner) {
+                        commandExecutionResults.push(kubectl.annotate('pod', pod.metadata.name, workflowAnnotations, true));
```
ds-ms, line 66: break after this maybe?

src/utilities/strategy-helpers/deployment-helper.ts
```diff
@@ -42,6 +45,16 @@ export async function deploy(kubectl: Kubectl, manifestFilePaths: string[], depl
...
+    
+    // annotate resources
+    let allPods: any;
+    try {
+        allPods = JSON.parse((kubectl.getAllPods()).stdout);
+    } catch (e) {
+        core.debug("Unable to parse pods; Error: " + e);
+    }
+
+    annotateResources(deployedManifestFiles, kubectl, resourceTypes, allPods);
```
ds-ms, line 57: Ensure failure in annotation doesn't break the workflow

ds-ms: Noticed the warnIfError part.. my bad

src/utilities/utility.ts
koushdey: refactored from `forEach` to `for..of` loop as break/return doesn't work.

src/utilities/utility.ts
koushdey: Done refactoring

src/utilities/utility.ts
koushdey: Done

PR #56, opened 2020-08-10
Annotation and Label changes to handle multiple workflows across branches

koushdey:
Annotation and Label changes to handle multiple unique workflows and parallel runs of those during new commits across branches. [ADR](https://github.com/github/Microsoft-Attach/pull/11/files)


ds-ms (changes requested):
Create a NEW branch for releases. call it `releases/v1.2`

ds-ms:
I'll review your changes in that PR.

PR #57, opened 2020-08-25
Annotation and Label changes to handle multiple workflows across branches

koushdey:
Annotation and Label changes to handle multiple unique workflows and parallel runs of those during new commits across branches. [ADR](https://github.com/github/Microsoft-Attach/pull/11/files)

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -25,11 +25,12 @@ export const deploymentTypes: string[] = ['deployment', 'replicaset', 'daemonset
 export const workloadTypes: string[] = ['deployment', 'replicaset', 'daemonset', 'pod', 'statefulset', 'job', 'cronjob'];
 export const workloadTypesWithRolloutStatus: string[] = ['deployment', 'daemonset', 'statefulset'];
 
-export function getWorkflowAnnotationsJson(lastSuccessRunSha: string): string {
+export function getWorkflowAnnotationsJson(lastSuccessRunSha: string, workflowFilePath: string): string {
     return `{`
         + `'run': '${process.env.GITHUB_RUN_ID}',`
         + `'repository': '${process.env.GITHUB_REPOSITORY}',`
         + `'workflow': '${process.env.GITHUB_WORKFLOW}',`
+        + `'workflowFileName': '${workflowFilePath.replace(".github/workflows/", "")}',`
         + `'jobName': '${process.env.GITHUB_JOB}',`
         + `'createdBy': '${process.env.GITHUB_ACTOR}',`
         + `'runUri': 'https://github.com/${process.env.GITHUB_REPOSITORY}/actions/runs/${process.env.GITHUB_RUN_ID}',`
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -40,9 +40,9 @@ export function getWorkflowAnnotationsJson(lastSuccessRunSha: string): string {
         + `'provider': 'GitHub'`
         + `}`;
 }
-export function getWorkflowAnnotationKeyLabel(): string {
+export function getWorkflowAnnotationKeyLabel(workflowFilePath: string): string {
     const hashKey = require("crypto").createHash("MD5")
-        .update(`${process.env.GITHUB_REPOSITORY}/${process.env.GITHUB_WORKFLOW}`)
+        .update(`${process.env.GITHUB_REPOSITORY}/${workflowFilePath}`)
         .digest("hex");
     return `githubWorkflow_${hashKey}`;
 }
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -25,15 +25,24 @@ export const deploymentTypes: string[] = ['deployment', 'replicaset', 'daemonset
 export const workloadTypes: string[] = ['deployment', 'replicaset', 'daemonset', 'pod', 'statefulset', 'job', 'cronjob'];
 export const workloadTypesWithRolloutStatus: string[] = ['deployment', 'daemonset', 'statefulset'];
 
-export const workflowAnnotations = [
-    `run=${process.env['GITHUB_RUN_ID']}`,
-    `repository=${process.env['GITHUB_REPOSITORY']}`,
-    `workflow=${process.env['GITHUB_WORKFLOW']}`,
-    `jobName=${process.env['GITHUB_JOB']}`,
-    `createdBy=${process.env['GITHUB_ACTOR']}`,
-    `runUri=https://github.com/${process.env['GITHUB_REPOSITORY']}/actions/runs/${process.env['GITHUB_RUN_ID']}`,
-    `commit=${process.env['GITHUB_SHA']}`,
-    `branch=${process.env['GITHUB_REF']}`,
-    `deployTimestamp=${Date.now()}`,
-    `provider=GitHub`
-];
+export function getWorkflowAnnotationsJson(lastSuccessRunSha: string): string {
+    return `{`
+        + `'run': '${process.env.GITHUB_RUN_ID}',`
+        + `'repository': '${process.env.GITHUB_REPOSITORY}',`
+        + `'workflow': '${process.env.GITHUB_WORKFLOW}',`
+        + `'jobName': '${process.env.GITHUB_JOB}',`
+        + `'createdBy': '${process.env.GITHUB_ACTOR}',`
+        + `'runUri': 'https://github.com/${process.env.GITHUB_REPOSITORY}/actions/runs/${process.env.GITHUB_RUN_ID}',`
+        + `'commit': '${process.env.GITHUB_SHA}',`
+        + `'lastSuccessRunCommit': '${lastSuccessRunSha}',`
+        + `'branch': '${process.env.GITHUB_REF}',`
+        + `'deployTimestamp': '${Date.now()}',`
+        + `'provider': 'GitHub'`
+        + `}`;
+}
+export function getWorkflowAnnotationKeyLabel(): string {
+    const hashKey = require("crypto").createHash("MD5")
+        .update(`${process.env.GITHUB_REPOSITORY}/${process.env.GITHUB_WORKFLOW}`)
+        .digest("hex");
+    return `githubWorkflow_${hashKey}`;
+}
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -40,6 +40,7 @@ export function getWorkflowAnnotationsJson(lastSuccessRunSha: string): string {
         + `'provider': 'GitHub'`
         + `}`;
 }
+
 export function getWorkflowAnnotationKeyLabel(workflowFilePath: string): string {
     const hashKey = require("crypto").createHash("MD5")
         .update(`${process.env.GITHUB_REPOSITORY}/${workflowFilePath}`)
--- a/src/githubClient.ts
+++ b/src/githubClient.ts
@@ -7,24 +7,20 @@ export class GitHubClient {
         this._token = token;
     }
 
-    public async getWorkflows(force?: boolean): Promise<any> {
-        if (force || !this._workflowsPromise) {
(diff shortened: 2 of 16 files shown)


src/constants.ts
```diff
@@ -25,15 +25,24 @@ export const deploymentTypes: string[] = ['deployment', 'replicaset', 'daemonset
...
+        + `'jobName': '${process.env.GITHUB_JOB}',`
+        + `'createdBy': '${process.env.GITHUB_ACTOR}',`
+        + `'runUri': 'https://github.com/${process.env.GITHUB_REPOSITORY}/actions/runs/${process.env.GITHUB_RUN_ID}',`
+        + `'commit': '${process.env.GITHUB_SHA}',`
+        + `'lastSuccessRunCommit': '${lastSuccessRunSha}',`
+        + `'branch': '${process.env.GITHUB_REF}',`
+        + `'deployTimestamp': '${Date.now()}',`
+        + `'provider': 'GitHub'`
+        + `}`;
+}
```
ds-ms, line 42: Add a newline here

src/kubectl-object-model.ts
```diff
@@ -65,6 +65,14 @@ export class Kubectl {
         return this.execute(args);
     }
 
+    public labelFiles(files: string | string[], labels: string[], overwrite?: boolean): IExecSyncResult {
+        let args = ['label'];
+        args = args.concat(['-f', this.createInlineArray(files)]);
+        args = args.concat(labels);
+        if (!!overwrite) { args.push(`--overwrite`); }
```
ds-ms, line 72: Is there a scenario which demands a non overwrite scenario?

src/utilities/utility.ts
```diff
@@ -50,50 +51,75 @@ export function checkForErrors(execResults: IExecSyncResult[], warnIfError?: boo
...
-                        commandExecutionResults.push(kubectl.annotate('pod', pod.metadata.name, workflowAnnotations, true));
+                        commandExecutionResults.push(kubectl.annotate('pod', pod.metadata.name, [annotationKeyValStr], true));
                     }
                 });
             }
         });
     }
-
     return commandExecutionResults;
 }
```
ds-ms, line 122: We're not annotating namespace anymore?

src/utilities/utility.ts
```diff
@@ -50,50 +51,75 @@ export function checkForErrors(execResults: IExecSyncResult[], warnIfError?: boo
...
+
+export function annotateChildPods(kubectl: Kubectl, resourceType: string, resourceName: string, annotationKeyValStr: string, allPods): IExecSyncResult[] {
     const commandExecutionResults = [];
     let owner = resourceName;
     if (resourceType.toLowerCase().indexOf('deployment') > -1) {
         owner = kubectl.getNewReplicaSet(resourceName);
     }
 
-    if (allPods && allPods.items && allPods.items.length > 0) {
+    if (!!allPods && !!allPods.items && allPods.items.length > 0) {
```
ds-ms, line 108: This `!!` addition is redundant

src/utilities/utility.ts
```diff
@@ -50,50 +51,75 @@ export function checkForErrors(execResults: IExecSyncResult[], warnIfError?: boo
...
+        }
+    }
+}
+
+export async function getWorkflowFilePath(githubToken: string): Promise<string> {
+    let workflowFilePath = process.env.GITHUB_WORKFLOW;
+    if (!workflowFilePath.startsWith('.github/workflows/')) {
+        const githubClient = new GitHubClient(process.env.GITHUB_REPOSITORY, githubToken);
+        const response = await githubClient.getWorkflows();
+        if (response.statusCode == StatusCodes.OK
```
ds-ms, line 82: add null check for response

src/utilities/utility.ts
ds-ms: It hinders readability.

src/utilities/utility.ts
```diff
@@ -50,50 +51,75 @@ export function checkForErrors(execResults: IExecSyncResult[], warnIfError?: boo
...
+            const annotationsSet = JSON.parse(result.stdout).metadata.annotations;
+            if (!!annotationsSet && !!annotationsSet[annotationKey]) {
+                return JSON.parse(annotationsSet[annotationKey].replace(/'/g, '"')).commit;
+            }
+            else {
+                return 'NA';
+            }
+        }
+    }
+}
```
ds-ms, line 75: consider removing `!!`

src/utilities/utility.ts
```diff
@@ -50,50 +51,75 @@ export function checkForErrors(execResults: IExecSyncResult[], warnIfError?: boo
...
+    }
+    else {
+        if (!!result.stderr) {
+            core.debug(`${result.stderr}`);
+            return process.env.GITHUB_SHA;
+        }
+        else if (!!result.stdout) {
+            const annotationsSet = JSON.parse(result.stdout).metadata.annotations;
+            if (!!annotationsSet && !!annotationsSet[annotationKey]) {
+                return JSON.parse(annotationsSet[annotationKey].replace(/'/g, '"')).commit;
```
ds-ms, line 68: Surround this with `try..catch`

src/utilities/utility.ts
```diff
@@ -50,50 +51,75 @@ export function checkForErrors(execResults: IExecSyncResult[], warnIfError?: boo
...
-    if (allPods && allPods.items && allPods.items.length > 0) {
+    if (!!allPods && !!allPods.items && allPods.items.length > 0) {
         allPods.items.forEach((pod) => {
             const owners = pod.metadata.ownerReferences;
-            if (owners) {
+            if (!!owners) {
                 owners.forEach(ownerRef => {
                     if (ownerRef.name === owner) {
-                        commandExecutionResults.push(kubectl.annotate('pod', pod.metadata.name, workflowAnnotations, true));
+                        commandExecutionResults.push(kubectl.annotate('pod', pod.metadata.name, [annotationKeyValStr], true));
```
ds-ms, line 114: Is it a single `annotation`? Or are you converting the array into a string and passing it here? 

If it's going to be a single string, consider changing the signature of the function

src/utilities/strategy-helpers/deployment-helper.ts
```diff
@@ -112,19 +112,36 @@ async function checkManifestStability(kubectl: Kubectl, resources: Resource[]):
...
+    labelResources(files, kubectl, annotationKeyLabel);
+}
+
+function annotateResources(files: string[], kubectl: Kubectl, resourceTypes: Resource[], allPods: any, annotationKey: string) {
     const annotateResults: IExecSyncResult[] = [];
-    annotateResults.push(annotateNamespace(kubectl, TaskInputParameters.namespace));
-    annotateResults.push(kubectl.annotateFiles(files, models.workflowAnnotations, true));
+    const lastSuccessSha = getLastSuccessfulRunSha(kubectl, TaskInputParameters.namespace, annotationKey);
+    let annotationKeyValStr = annotationKey + '=' + models.getWorkflowAnnotationsJson(lastSuccessSha);
+    annotateResults.push(kubectl.annotate('namespace', TaskInputParameters.namespace, [annotationKeyValStr], true));
```
ds-ms, line 126: Is it going to be a single annotation from now on?

src/utilities/strategy-helpers/deployment-helper.ts
```diff
@@ -112,19 +112,36 @@ async function checkManifestStability(kubectl: Kubectl, resources: Resource[]):
...
         }
     });
     checkForErrors(annotateResults, true);
 }
 
+function labelResources(files: string[], kubectl: Kubectl, label: string) {
+    let workflowName = process.env.GITHUB_WORKFLOW;
+    workflowName = workflowName.startsWith('.github/workflows/') ?
+        workflowName.replace(".github/workflows/", "") : workflowName;
+    const labels = [`workflowFriendlyName=${workflowName}`, `workflow=${label}`];
```
ds-ms, line 141: Why the camel casing?

src/githubClient.ts
```diff
@@ -0,0 +1,30 @@
...
+import { WebRequest, WebResponse, sendRequest } from "./utilities/httpClient";
+
+export class GitHubClient {
+    constructor(repository: string, token: string) {
+        this._repository = repository;
+        this._token = token;
+    }
+
+    public async getWorkflows(force?: boolean): Promise<any> {
+        if (force || !this._workflowsPromise) {
```
ds-ms, line 11: I don't see a need for caching given that it gets called from a single place.

koushdey: Added it, keeping in mind if we need to call API from multiple places. Want me to remove it?

src/kubectl-object-model.ts
koushdey: There is no scenario currently with no overwrite. Kept it aligning to `annotate` method, for future scenarios. Want me to remove it?

src/utilities/strategy-helpers/deployment-helper.ts
koushdey: Yes. In this flow, it is going to be a single key with a JSON Object (string ) as value.

src/utilities/strategy-helpers/deployment-helper.ts
koushdey: That has been suggested by Rachit to be part of the label selector. Any suggestions?

src/utilities/utility.ts
koushdey: Done

src/utilities/utility.ts
koushdey: Done

src/utilities/utility.ts
koushdey: Done

src/utilities/utility.ts
koushdey: Done

src/utilities/utility.ts
koushdey: It is a single key with JSON object as value. Array is not converted into string here. Kept the signature as array to felicitate multiple annotations for future scenarios. Want me to remove it?

src/utilities/utility.ts
koushdey: we are still annotating namespace in `annotateResources` method in `deployment-helper`. It has been removed from here because the annotate namespace is no longer conditional.

src/kubectl-object-model.ts
ds-ms: Yep.

src/utilities/utility.ts
ds-ms: Yes, the current scenario clearly doesn't warrant this.

src/utilities/strategy-helpers/deployment-helper.ts
ds-ms: `workflow-friendly-name` abiding to k8s casing

src/githubClient.ts
ds-ms: Remove it, if this flow is not being hit. :)

ds-ms (approved):
LGTM!
I'll let you take a call on the comments I gave 😄

src/utilities/strategy-helpers/deployment-helper.ts
```diff
@@ -115,14 +115,14 @@ async function checkManifestStability(kubectl: Kubectl, resources: Resource[]):
 async function annotateAndLabelResources(files: string[], kubectl: Kubectl, resourceTypes: Resource[], allPods: any) {
     const workflowFilePath = await getWorkflowFilePath(TaskInputParameters.githubToken);
```
ds-ms, line 116: Is the workflow expected to fail if any of these extra HTTP calls fail?

koushdey: No. If the HTTP call fails it will go ahead with the value of` GITHUB_WORKFLOW` variable

PR #64, opened 2020-09-21
(M)Annotation and Label changes to handle multiple workflows across branches

koushdey:
(Master)Annotation and Label changes to handle multiple workflows across branches


ajinkya599 (approved):
- Assuming code same as in releases/v1.2.
- Okay to ignore failing blue-green integration tests on windows/macOS (Passing on ubuntu. Need to fix tests on windows)

Issue #34, opened 2020-05-27
Add ability to apply whole directories


dberzano:
Same as:

```
kubectl apply -f my/dir/with/yaml/files/
```

Thanks!

ds-ms:
cc: @anraghun 

I feel it's cleaner if we introduce a wild character like input where you supply something like `my/dir/with/yaml/files/*.yml` or something like that? Then it shouldn't restrict you from having other types of files in the folder

ds-ms:
@anraghun any updates on the decision as to what we should go forward with?

anraghun:
@ds-ms I like the approach you have proposed. We need to have a file based filter otherwise we would be mandating adding only manifest files to a particular folder. 
Let's proceed with `my/dir/with/yaml/files/*.yml`

Enemico:
We also would really appreciate this feature.

ds-ms:
@shigupt202 could you pick this up?

shigupt202:
This feature can be accessed in v1.4 release of the action

apgapg:
@shigupt202 Any sample docs on implementing this?
I don't see how can i implement above thing.

PR #69, opened 2020-10-27
Adding logic to resolve glob patterns in file path

shigupt202:
Feature Request: https://github.com/Azure/k8s-deploy/issues/34


ds-ms:
This will have be raised for a new `releases/v*` branch

PR #70, opened 2020-10-29
Adding logic for glob in file paths

shigupt202:


src/utilities/utility.ts
```diff
@@ -140,3 +141,24 @@ export function getRandomInt(max: number) {
...
+        core.setFailed('No manifests supplied to deploy');
+        return;
+    }
+    let unresolvedManifests = manifests.split('\n');
+    let resolvedManifests: string[] = [];
+    unresolvedManifests.forEach((manifestPath) => {
+        if (glob.hasMagic(manifestPath))
+            resolvedManifests = resolvedManifests.concat(glob.sync(manifestPath));
+        else
+            resolvedManifests.push(manifestPath);
```
ds-ms, line 156: Surround logic in curly braces, or have it inline.

src/utilities/utility.ts
```diff
@@ -140,3 +141,24 @@ export function getRandomInt(max: number) {
...
+    });
+
+    if (resolvedManifests == null || resolvedManifests.length == 0) {
```
ds-ms, line 159: resolvedManifests will never be null. 

You could also just change it to `if (resolvedManifests)`

ds-ms (approved):
LGTM. 

Add a test case to our L2s while before merging this PR.

Issue #63, opened 2020-09-18
[Blue-Green Bug Bash] Ingress promote/reject modifies non blue-green ingresses


ajinkya599:
Repro steps:
- Deploy with `route-method` as `ingress`
- Promote the deployment

Current behaviour:
- On promote, the ingress which is not a part of blue/green also gets applied [Internal run id: 1134759256]
![image](https://user-images.githubusercontent.com/11447401/93622249-6d5a5c00-f9fa-11ea-832a-3992d2764f9f.png)

Expected behaviour:
- Only the ingresses that are identified for blue/green should be applied.

ajinkya599:
Same is true for reject:
![image](https://user-images.githubusercontent.com/11447401/93623023-b65ee000-f9fb-11ea-8ad6-f8d75f2f0afa.png)

Issue #60, opened 2020-09-17
[Blue-Green Bug Bash] Doc update for blue green startegy


shigupt202:
The README should be updated with `route-method: ingress` instead of `strategy: ingress` [here](https://github.com/Azure/k8s-deploy/blob/b824a8e3caa166c049c7c83e26c11af1b242b937/README.md#to-promotereject-the-green-workload-created-by-the-above-snippet-the-following-yaml-snippet-could-be-used)

PR #71, opened 2020-10-29
Fixing issues found in bug bash.

### LINKED ISSUE #59: [Blue-Green Bug Bash] Promote SMI deletes workloads first and then the TrafficSplit
Repro steps:
- Deploy with `route-method` as `smi`
- Promote the deployment

Current behaviour:
- Deployment and temporary services are deleted first, and then the TrafficSplit resource is deleted
image

Expected behaviour:
- TrafficSplit should be deleted first. Otherwise, during the transition, there would be broken references of services in TrafficSplit

ajinkya599:
Fixes #59, #60 and #63:

- Update readme
- Promote/reject ingress - Apply only the modified ingresses (which are part of blue-green)
Before:
image
After:
image

- Promote/reject smi - Delete trafficSplit before deleting the backend services
Before:
image
After:
image


ds-ms (approved):
LGTM

shigupt202 (approved):

PR #73, opened 2020-10-31
Fixing issues found in bug bash. (#71)

ajinkya599:
Porting changes from #71

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,7 +218,7 @@ Following are the key capabilities of this action:
         service.yaml
         ingress-yml
     strategy: blue-green
-    strategy: ingress # should be the same as the value when action was deploy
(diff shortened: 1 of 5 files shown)


ds-ms (approved):

shigupt202:
Shouldn't we create new release versions for new changes? Or since it's bug fix, it's ok to push them to older versions?

ds-ms:
I approved cause it was a very minimal bug fix. 
We should release a new tag `v1.3.1` though.

shigupt202 (approved):

ajinkya599:
tag `v1.3` does not exist. So, I think we should release this as `v1.3`.

PR #74, opened 2020-11-01
Adding repositoryId to the list of annotations

rachitb777:


src/utilities/utility.ts
```diff
@@ -105,6 +105,24 @@ export async function getWorkflowFilePath(githubToken: string): Promise<string>
     return Promise.resolve(workflowFilePath);
 }
 
+export async function getRepositoryId(githubToken: string): Promise<number> {
+    const githubClient = new GitHubClient(process.env.GITHUB_REPOSITORY, githubToken);
+    let repoId = 0;
+    const response = await githubClient.getRepo();
+    if (response) {
+        if (response.statusCode == StatusCodes.OK
+            && response.body) {
```
ds-ms, line 114: Don't you need to handle the case where `response.body` is null and StatusCode is OK.

src/utilities/utility.ts
```diff
@@ -105,6 +105,24 @@ export async function getWorkflowFilePath(githubToken: string): Promise<string>
...
+                repoId = response.body.id;
+        }
+        else if (response.statusCode != StatusCodes.OK) {
+            core.debug(`An error occured while getting repo. Statuscode: ${response.statusCode}, StatusMessage: ${response.statusMessage}`);
+        }
+    } else {
+        core.warning(`Failed to get response from repos GET API`);
+    }
+    return Promise.resolve(repoId);
```
ds-ms, line 123: Just return `repoId`

src/utilities/utility.ts
```diff
@@ -105,6 +105,24 @@ export async function getWorkflowFilePath(githubToken: string): Promise<string>
     return Promise.resolve(workflowFilePath);
 }
 
+export async function getRepositoryId(githubToken: string): Promise<number> {
```
ds-ms, line 108: Set the return type to `number`.
This way your function would return a `Promise<number>` instead of `Promise<Promise<number>>` because of the async wrapping

src/utilities/strategy-helpers/deployment-helper.ts
```diff
@@ -129,16 +129,17 @@ async function checkManifestStability(kubectl: Kubectl, resources: Resource[]):
 }
 
 async function annotateAndLabelResources(files: string[], kubectl: Kubectl, resourceTypes: Resource[], allPods: any) {
+    const repositoryId = await getRepositoryId(TaskInputParameters.githubToken);
```
ds-ms, line 132: The value here would be a Promise right?

ds-ms (changes requested):

src/utilities/strategy-helpers/deployment-helper.ts
rachitb777: we are using async/await pattern. Since we are using await on a Promise, the return value is of the type number.

src/utilities/utility.ts
rachitb777: I think it wraps only if the Promise is missing. I would like to be consistent with rest of the code.

src/utilities/utility.ts
rachitb777: Will do.

src/utilities/utility.ts
rachitb777: 👍

src/utilities/strategy-helpers/deployment-helper.ts
ds-ms: No no, I said this because I noticed your return type to be `Promise<number>`

ds-ms:
You need to update the tests as well to make the L0s green

ds-ms (approved):
The change LGTM 👍 

Could you add a test for this as well :)

ds-ms:
@rachitb777 pinging to take this PR forward.

ds-ms:
Closing this PR @rachitb777 feel free to reopen it if this scenario is still valid.

PR #77, opened 2020-11-12
Dockerfile, manifest and helm file annotations

Josh-01:
Added change to include following to annotations:

1. Manifest files path
2. Helm chart path
3. For each image:
    - Dockerfile path
    - CR server name


Josh-01:
Wrong branches, hence closing

PR #82, opened 2020-11-26
Updated new features in Readme with sample workflows

Josh-01:
Added workflow samples for using the new env variables to enable traceability features for dockerfile, manifests, helmcharts and acr.

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,16 +318,16 @@ jobs:
 
 ### End to end workflow for building and deploying container images 
 
- - Env variables `CR_USERNAME` and `CR_PASSWORD` are for container registry login used in various places in the workflow
- - Env variable `DOCKERFILE_PATHS` is a list of comma separated pairs of (image-name, dockerfile-path)
+ - Env variable `REGISTRY_CREDENTIALS` list of credentials for container registry login as <registry_name><space><username><space><password>
+ - Env variable `DOCKERFILE_PATHS` is a list of dockerfile paths for images used as <image_name><space><dockerfile_path>
 
 ```yaml
 on: [push]
 env:
   REGISTRY_URL: contoso.azurecr.io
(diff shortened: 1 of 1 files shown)


Josh-01:
V1.2 has been updated, so the new features are available in the action. I can now merge this PR once it is approved, so that users can see the new env-variables/ labels expected for traceability in the README.

koushdey (approved):
Check on the integration test failures. I think those are not because of you.

README.md
```diff
@@ -314,6 +314,139 @@ jobs:
...
+
+```yaml
+on: [push]
+env:
+  REGISTRY_URL: contoso.azurecr.io
+  NAMESPACE: testnamespace1
+  REGISTRY_CREDENTIALS: |
+    contoso.azurecr.io ${ secrets.AZURE_CLIENT_ID } ${ secrets.AZURE_CLIENT_SECRET }
+  DOCKERFILE_PATHS: |
+    contoso.azurecr.io/k8sdemo:first ./Dockerfile
```
ds-ms, line 332: I don't think this is a good design. 

Also take this change along with the PR.

Josh-01: I agree it was bad design - I have reverted this particular change. We have already pushed the current changes to v1.2, so we should atleast update the README to reflect those changes - this is long overdue.
I will raise a new set of PRs with the design changes required and all other review comment changes, for master(with updated README), v1.2,v1.3 and v1.4.

ds-ms: Why did you merge this PR? Given that you were changing the logic anyway?

Josh-01: Like I mentioned in my above comment while merging, the changes had been discussed and pushed to v1.2 in code, and only the README update was pending for a long time. After discussing with Ambika, he suggested we go ahead with the merge and add the new changes as an incremental update, since it was pending discussion.

PR #78, opened 2020-11-17
Initial commit for config files annotations

Josh-01:
Added change to include following to annotations:

1. Manifest files path
2. Helm chart path
3. For each image:
   - Dockerfile path
   - CR server name

Issues: https://github.com/github/GitHub-EcoSystem/issues/690

PR #83, opened 2020-11-26
Adding annotation new changes to v1.3

Josh-01:
Similar to https://github.com/Azure/k8s-deploy/pull/78 for v1.2.

Additional changes:
Design to handle multiple registries and images updated
Will update the same in version 1.2 and master.


src/constants.ts
```diff
@@ -25,21 +25,25 @@ export const deploymentTypes: string[] = ['deployment', 'replicaset', 'daemonset
...
+    annotationObject["createdBy"] = process.env.GITHUB_ACTOR;
+    annotationObject["runUri"] = `https://github.com/${process.env.GITHUB_REPOSITORY}/actions/runs/${process.env.GITHUB_RUN_ID}`;
+    annotationObject["commit"] = process.env.GITHUB_SHA;
+    annotationObject["lastSuccessRunCommit"] = lastSuccessRunSha;
+    annotationObject["branch"] = process.env.GITHUB_REF;
+    annotationObject["deployTimestamp"] = Date.now();
+    annotationObject["dockerfilePaths"] = filePathConfigs.dockerfilePaths;
+    annotationObject["manifestsPaths"] = filePathConfigs.manifestFilePaths
+    annotationObject["helmChartPaths"] = filePathConfigs.helmChartFilePaths;
+    annotationObject["provider"] = "GitHub";
```
ds-ms, line 44: You could have created the object in place right? like
`let annotationObject: any = {
  "run": `${process.env.GITHUB_RUN_ID}`,
...
}`

src/utilities/utility.ts
```diff
@@ -129,6 +131,93 @@ export function annotateChildPods(kubectl: Kubectl, resourceType: string, resour
...
+    let filePathsConfig: any = {};
+    const MANIFEST_PATHS_KEY = 'manifestFilePaths';
+    const HELM_CHART_KEY = 'helmChartFilePaths';
+    const DOCKERFILE_PATH_KEY = 'dockerfilePaths';
+    const DOCKERFILE_PATH_LABEL_KEY = 'dockerfile-path';
+
+    let inputManifestFiles = inputParams.manifests || [];
+    filePathsConfig[MANIFEST_PATHS_KEY] = inputManifestFiles;
+
+    let helmChartPaths = (process.env.HELM_CHART_PATHS && process.env.HELM_CHART_PATHS.split('\n').filter(path => path != "")) || [];
```
ds-ms, line 145: Where is this value coming from? `k8s-bake`?

src/utilities/utility.ts
```diff
@@ -129,6 +131,93 @@ export function annotateChildPods(kubectl: Kubectl, resourceType: string, resour
...
+    filePathsConfig[HELM_CHART_KEY] = helmChartPaths;
+
+    //Parsing dockerfile paths for images
+    let imageNames = core.getInput('images').split('\n');
+    let imageDockerfilePathMap: any = {};
+    let pathKey: any, pathVal: any;
+
+    //Fetching from env var if available
+    let dockerfilePathsList: any[] = (process.env.DOCKERFILE_PATHS &&  process.env.DOCKERFILE_PATHS.split('\n')) || [];
```
ds-ms, line 154: Who's expected to set this value?

src/utilities/utility.ts
```diff
@@ -129,6 +131,93 @@ export function annotateChildPods(kubectl: Kubectl, resourceType: string, resour
...
+    dockerfilePathsList.forEach(path => {
+        if(path){
+            pathKey = path.split(' ')[0];
+            pathVal = path.split(' ')[1];
+            imageDockerfilePathMap[pathKey] = pathVal;
+        }
+    })
+
+    //Fetching from image lable if available
+    for(const image of imageNames){
```
ds-ms, line 164: indent

ds-ms: Run a file formatter once

src/utilities/utility.ts
```diff
@@ -129,6 +131,93 @@ export function annotateChildPods(kubectl: Kubectl, resourceType: string, resour
...
+        let args: string[] = [image];
+        let resultObj: any;
```
ds-ms, line 166: Just use `result`

src/utilities/utility.ts
```diff
@@ -129,6 +131,93 @@ export function annotateChildPods(kubectl: Kubectl, resourceType: string, resour
...
+        let containerRegistryName = image;
+
+        try{
+            let usrname = process.env.CR_USERNAME || null;
```
ds-ms, line 170: `username`

src/utilities/utility.ts
```diff
@@ -129,6 +131,93 @@ export function annotateChildPods(kubectl: Kubectl, resourceType: string, resour
...
+            let pwd = process.env.CR_PASSWORD || null;
```
ds-ms, line 171: `password`

ds-ms: You could also change this to registryUsername and registryPassword

src/utilities/utility.ts
```diff
@@ -129,6 +131,93 @@ export function annotateChildPods(kubectl: Kubectl, resourceType: string, resour
...
+            await exec.exec('docker inspect --type=image', args, true).then(res => {
+                if (res.stderr != '' && !res.success) {
+                    throw new Error(`docker inspect call failed with: ${res.stderr.match(/(.*)\s*$/)![0]}`);
+                }
+
+                if(res.stdout){
+                    resultObj = JSON.parse(res.stdout);
+                }
+            });   
+        }
```
ds-ms, line 197: How are you handling a multi image scenario?

src/utilities/utility.ts
```diff
@@ -129,6 +131,93 @@ export function annotateChildPods(kubectl: Kubectl, resourceType: string, resour
...
+        catch (ex) {
+            core.warning(`Failed to get dockerfile paths for image ${image.toString()} | ` + ex);
+        }
+
+        if(resultObj){
```
ds-ms, line 202: What's the point of `result` if you're not returning it. The variable name doesn't justify the usage in the flow

src/utilities/utility.ts
```diff
@@ -129,6 +131,93 @@ export function annotateChildPods(kubectl: Kubectl, resourceType: string, resour
     return commandExecutionResults;
 }
 
+export async function getFilePathsConfigs(): Promise<any> {
```
ds-ms, line 134: Have an interface for the result. Instead of returning any

src/utilities/utility.ts
```diff
@@ -129,6 +131,93 @@ export function annotateChildPods(kubectl: Kubectl, resourceType: string, resour
...
+            });   
+        }
+        catch (ex) {
+            core.warning(`Failed to get dockerfile paths for image ${image.toString()} | ` + ex);
+        }
+
+        if(resultObj){
+            resultObj = resultObj[0];
+            if((resultObj.Config) && (resultObj.Config.Labels) && (resultObj.Config.Labels[DOCKERFILE_PATH_LABEL_KEY])){
+                pathVal = resultObj.Config.Labels[DOCKERFILE_PATH_LABEL_KEY];
```
ds-ms, line 205: Don't use short forms

src/constants.ts
Josh-01: Creating it in place was causing parsing errors when the values are arrays or objects. This way the object gets correctly parsed and back using JSON stringify and parse.

src/utilities/utility.ts
Josh-01: No it is an environment variable in the workflow, the path to the helmchart, that is also used by the ks8-bake action.

src/utilities/utility.ts
Josh-01: The user would be setting it if required. It is explained in the updated README with workflow samples in the other PR.

src/utilities/utility.ts
Josh-01: Done

src/utilities/utility.ts
Josh-01: Done

src/utilities/utility.ts
Josh-01: Done

src/utilities/utility.ts
Josh-01: Done

src/utilities/utility.ts
Josh-01: Done

src/utilities/utility.ts
Josh-01: Done

src/utilities/utility.ts
Josh-01: It is not handled here.
I will make the env variable an array of values that would be parsed, to handle the multiple images from multiple registries.

src/utilities/utility.ts
Josh-01: Done

__tests__/run.test.ts
```diff
@@ -1,393 +1,396 @@
-import * as KubernetesManifestUtility from '../src/utilities/manifest-stability-utility';
```
koushdey, line 1: Fix LF/CRLF for this file.

lib/utilities/utility.js
```diff
@@ -1,242 +1,248 @@
...
+}
+exports.sleep = sleep;
+function getRandomInt(max) {
+    return Math.floor(Math.random() * Math.floor(max));
+}
+exports.getRandomInt = getRandomInt;
+function getCurrentTime() {
+    return new Date().getTime();
+}
+exports.getCurrentTime = getCurrentTime;
```
ds-ms: Use `npm run-script build`

__tests__/run.test.ts
Josh-01: Done

lib/utilities/utility.js
Josh-01: Done

src/utilities/utility.ts
Josh-01: Added the design change for multiple registries-credentials and multiple image-dockerfile paths.

src/utilities/utility.ts
ds-ms: Did you callout these assumptions explicitly in the README?

src/utilities/exec.ts
```diff
@@ -0,0 +1,33 @@
+import * as aexec from '@actions/exec';
+
+export interface ExecResult {
+  success: boolean;
+  stdout: string;
+  stderr: string;
+}
+
+export const exec = async (command: string, args: string[] = [], silent?: boolean): Promise<ExecResult> => {
```
ds-ms, line 9: Consider using `ToolRunner` instead of `exec`

src/utilities/utility.ts
```diff
@@ -129,6 +137,75 @@ export function annotateChildPods(kubectl: Kubectl, resourceType: string, resour
...
+    //Fetching from image label if available
+    for (const image of imageNames) {
+        let args: string[] = [image];
+        let imageConfig: any;
+
+        try {
+            
+            await exec.exec('docker pull ', args, true).then(res => {
+                if (res.stderr != '' && !res.success) {
+                    throw new Error(`docker images pull failed with: ${res.stderr.match(/(.*)\s*$/)![0]}`);
```
ds-ms, line 168: Is the flow expected to fail, if `docker pull` fails? 
I noticed you're relying on the `system` handling for errors. Should you just catch and display a warning instead?

src/utilities/utility.ts
```diff
@@ -129,6 +137,75 @@ export function annotateChildPods(kubectl: Kubectl, resourceType: string, resour
...
+                }
+            });
+
+            await exec.exec('docker inspect --type=image', args, true).then(res => {
```
ds-ms, line 172: have a docker object model, similar to [k8s-object-model](https://github.com/Azure/k8s-deploy/blob/master/src/kubectl-object-model.ts). 
That way you could invoke and do the explicit parsings you need.

src/utilities/utility.ts
```diff
@@ -129,6 +137,75 @@ export function annotateChildPods(kubectl: Kubectl, resourceType: string, resour
...
+                pathValue = 'Not available';
+            }
+                
+            imageDockerfilePathMap[image] = pathValue;
+        }
+    }
+
+    filePathsConfig[DOCKERFILE_PATH_KEY] = imageDockerfilePathMap;
+
+    return Promise.resolve(filePathsConfig);
```
ds-ms, line 206: Just return `filePathsConfig`

src/utilities/utility.ts
```diff
@@ -129,6 +137,75 @@ export function annotateChildPods(kubectl: Kubectl, resourceType: string, resour
...
+        catch (ex) {
+            core.warning(`Failed to get dockerfile paths for image ${image.toString()} | ` + ex);
+        }
+
+        if (imageConfig) {
+            imageConfig = imageConfig[0];
+            if ((imageConfig.Config) && (imageConfig.Config.Labels) && (imageConfig.Config.Labels[DOCKERFILE_PATH_LABEL_KEY])) {
+                pathValue = imageConfig.Config.Labels[DOCKERFILE_PATH_LABEL_KEY];
+                if(pathValue.startsWith('./')){  //if it is relative filepath convert to link from current repo
+                    pathLink = `https://github.com/${process.env.GITHUB_REPOSITORY}/${branchOrTag}/${pathValue}`;
```
ds-ms, line 191: What if the image is built in a different repo?

src/utilities/utility.ts
```diff
@@ -129,6 +137,75 @@ export function annotateChildPods(kubectl: Kubectl, resourceType: string, resour
...
+    let inputManifestFiles = inputParams.manifests || [];
+    filePathsConfig[MANIFEST_PATHS_KEY] = inputManifestFiles;
+
+    let helmChartPaths = (process.env.HELM_CHART_PATHS && process.env.HELM_CHART_PATHS.split('\n').filter(path => path != "")) || [];
+    filePathsConfig[HELM_CHART_KEY] = helmChartPaths;
+
+    let imageNames = core.getInput('images').split('\n');
+    let imageDockerfilePathMap: any = {};
+    let pathValue: string, pathLink: string;
+    const branchOrTag: string = process.env.GITHUB_REF && process.env.GITHUB_REF.replace('refs/heads/','blob/').replace('refs/tags/','blob/');
```
ds-ms, line 157: What value does `blob/` do? is it for creating a path to a particular file? If so, you should hard-code the constant where you're generating it and let the `branchOrTag` be set to the particular name. 

Also consider calling the variable `reference` maybe? It seems a little generic

src/utilities/utility.ts
```diff
@@ -129,6 +137,75 @@ export function annotateChildPods(kubectl: Kubectl, resourceType: string, resour
...
+
+    let inputManifestFiles = inputParams.manifests || [];
+    filePathsConfig[MANIFEST_PATHS_KEY] = inputManifestFiles;
+
+    let helmChartPaths = (process.env.HELM_CHART_PATHS && process.env.HELM_CHART_PATHS.split('\n').filter(path => path != "")) || [];
+    filePathsConfig[HELM_CHART_KEY] = helmChartPaths;
+
+    let imageNames = core.getInput('images').split('\n');
+    let imageDockerfilePathMap: any = {};
+    let pathValue: string, pathLink: string;
```
ds-ms, line 156: what path value?

ds-ms (changes requested):

src/utilities/utility.ts
Josh-01: Yes, it is mentioned in the README.

src/utilities/utility.ts
Josh-01: The path to the dockerfile shows up as :
https://github.com/<Username/RepoName>/blob/<branchOrTagName>/<DockerfileName>
The value is either the branch or tag name , hence the naming. 

`blob/` is part of the filepath. Is there another better way to get the filepath that I missed ?

src/utilities/utility.ts
Josh-01: No it is throwing an error if pull fails and another if inspect fails (to show exactly why it failed), and then it is caught and warning it displayed in the next few lines, like you have suggested. Did I understand correctly?

src/utilities/utility.ts
Josh-01: I did see the k8s object model, but that is used in multiple files and so it is useful.
The docker command is used only for these 2 calls - does it make sense to create another object model for it?

src/utilities/utility.ts
Josh-01: I have mentioned this in the README : 
`Use script to build image and add `dockerfile-path` label to it. The value expected is the link to the dockerfile. 
   If your dockerfile is in the same repo and branch where the workflow is run, it can be a relative path and it will be converted to a link for traceability.`

So, the the user is expected to give the correct file link if image is built in another repo.

src/utilities/utility.ts
Josh-01: It is an async call that expects promise, since we are calling container registries via docker. It was throwing errors if I make it sync - why do we need to make it sync ?

src/utilities/utility.ts
Josh-01: The variables are used in the loop to populate dockerfile paths and links.

src/utilities/utility.ts
ds-ms: Yeah it does, it will be like a single place to administer your docker calls. That way you can copy paste it in our different actions too, if needed

src/utilities/utility.ts
ds-ms, line 191: What I meant was, use `/blob/${branchOrTag}` instead, here.

ds-ms: Ok. I don't think this approach is user-friendly. But if that's what it takes to ensure e2e traceability so be it!

src/utilities/utility.ts
ds-ms: What error was it throwing? 

When you return `Promise<resolve>`, the expected js would be `Promise<Promise<resolve>>` right?

src/utilities/utility.ts
ds-ms: I think I missed the catch clause. My bad.

src/utilities/utility.ts
ds-ms: Addressed it in a comment below

ds-ms (approved):
Approving this PR to unblock you

src/utilities/utility.ts
Josh-01: Ok - resolving this then.

src/utilities/utility.ts
Josh-01: I agree, it is another step for the user, but for now (without something like Evidence store) we have to make do with this.

src/utilities/utility.ts
Josh-01: ok, resolving this.

src/utilities/utility.ts
Josh-01: Got it, will edit this.

src/utilities/utility.ts
Josh-01: No, it will return a Promise<FileConfigPath> object , which is what is expected from the await call to the function

src/utilities/exec.ts
Josh-01: Made the required changes for this

src/utilities/utility.ts
Josh-01: Ok, added the required changes for this.

Josh-01:
@koushdey @ajinkya599  Could you please review this? I have added the review changes discussed with @ds-ms.

__tests__/run.test.ts
koushdey: Seems like it is still not fixed

lib/constants.js
```diff
@@ -1,51 +1,54 @@
-'use strict';
-Object.defineProperty(exports, "__esModule", { value: true });
-exports.getWorkflowAnnotationKeyLabel = exports.getWorkflowAnnotationsJson = exports.workloadTypesWithRolloutStatus = exports.workloadTypes = exports.deploymentTypes = exports.ServiceTypes = exports.DiscoveryAndLoadBalancerResource = exports.KubernetesWorkload = void 0;
-class KubernetesWorkload {
-}
```
koushdey, line 5: Check for LF/CRLF here as well.

koushdey (approved):

lib/constants.js
Josh-01: Done

src/utilities/utility.ts
```diff
@@ -4,6 +4,18 @@ import { IExecSyncResult } from './tool-runner';
 import { Kubectl } from '../kubectl-object-model';
 import { GitHubClient } from '../githubClient';
 import { StatusCodes } from "./httpClient";
+import * as inputParams from "../input-parameters";
+import { DockerExec } from '../docker-object-model';
+import * as io from '@actions/io';
+import * as path from 'path';
```
ajinkya599, line 10: Remove unused dependencies.

src/docker-object-model.ts
```diff
@@ -0,0 +1,33 @@
+import { ToolRunner, IExecOptions, IExecSyncResult } from "./utilities/tool-runner";
+
+export class DockerExec{
+    private dockerPath: string;
+
+    constructor(dockerPath: string){
+        this.dockerPath = dockerPath;
+    };
+
+    public pullImage(args: string[], silent?: boolean) {
```
ajinkya599, line 10: - Rename `pullImage` -> `pull` and `inspectImage` -> `inspect`?
- Should these methods have signature like `pull(image: string, args: string[], silent?: boolean)` as `image` is always required?

src/docker-object-model.ts
```diff
@@ -0,0 +1,33 @@
...
+        args = ['pull', ...args];
+        var result: IExecSyncResult = this.execute(args,silent);
+        if (result.stderr != '' && result.code != 0) {
+            throw new Error(`docker images pull failed with: ${result.error}`);
+        }
+    }
+
+    public inspectImage(args: string[], silent?: boolean): any {
+        args = ['inspect', ...args];
+        var result: IExecSyncResult = this.execute(args,silent);
```
ajinkya599, line 20: Prefer `let` over `var`.
Prefer `const` over `let` for read-only properties.

src/constants.ts
```diff
@@ -25,21 +26,25 @@ export const deploymentTypes: string[] = ['deployment', 'replicaset', 'daemonset
...
+    annotationObject["repository"] = process.env.GITHUB_REPOSITORY;
+    annotationObject["workflow"] = process.env.GITHUB_WORKFLOW;
+    annotationObject["workflowFileName"] = workflowFilePath.replace(".github/workflows/", "");
+    annotationObject["jobName"] = process.env.GITHUB_JOB;
+    annotationObject["createdBy"] = process.env.GITHUB_ACTOR;
+    annotationObject["runUri"] = `https://github.com/${process.env.GITHUB_REPOSITORY}/actions/runs/${process.env.GITHUB_RUN_ID}`;
+    annotationObject["commit"] = process.env.GITHUB_SHA;
+    annotationObject["lastSuccessRunCommit"] = lastSuccessRunSha;
+    annotationObject["branch"] = process.env.GITHUB_REF;
+    annotationObject["deployTimestamp"] = Date.now();
```
ajinkya599, line 41: <ins>Discussion not related to this change</ins>:

Should we use new `new Date().toISOString();` over `Date.now()` as its more readable?
`Date.now()` -> `1609236761701` (epochs)
`new Date().toISOString();` -> `2020-12-29T10:12:35.773Z` (UTC)

@koushdey - Will it break anything down the line?

src/docker-object-model.ts
```diff
@@ -0,0 +1,33 @@
+import { ToolRunner, IExecOptions, IExecSyncResult } from "./utilities/tool-runner";
+
+export class DockerExec{
```
ajinkya599, line 3: Recommend running some formatting tool for all the files.

ajinkya599 (changes requested):

src/utilities/utility.ts
```diff
@@ -4,6 +4,18 @@ import { IExecSyncResult } from './tool-runner';
...
 import { GitHubClient } from '../githubClient';
 import { StatusCodes } from "./httpClient";
+import * as inputParams from "../input-parameters";
+import { DockerExec } from '../docker-object-model';
+import * as io from '@actions/io';
+import * as path from 'path';
+import * as toolCache from '@actions/tool-cache';
+
+
+export interface FileConfigPath {
```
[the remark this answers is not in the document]
Josh-01: Done

src/utilities/utility.ts
```diff
@@ -129,6 +141,62 @@ export function annotateChildPods(kubectl: Kubectl, resourceType: string, resour
     return commandExecutionResults;
 }
 
+export async function getFilePathsConfigs(): Promise<FileConfigPath> {
+
```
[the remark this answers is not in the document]
Josh-01: done

src/utilities/utility.ts
```diff
@@ -129,6 +141,62 @@ export function annotateChildPods(kubectl: Kubectl, resourceType: string, resour
     return commandExecutionResults;
 }
 
+export async function getFilePathsConfigs(): Promise<FileConfigPath> {
+
+    let filePathsConfig = <FileConfigPath>{};
```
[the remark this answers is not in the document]
Josh-01: Done

src/utilities/utility.ts
Josh-01: Done

src/utilities/utility.ts
```diff
@@ -129,6 +141,62 @@ export function annotateChildPods(kubectl: Kubectl, resourceType: string, resour
     return commandExecutionResults;
 }
 
+export async function getFilePathsConfigs(): Promise<FileConfigPath> {
```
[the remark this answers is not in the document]
Josh-01: Done

src/utilities/utility.ts
```diff
@@ -129,6 +141,62 @@ export function annotateChildPods(kubectl: Kubectl, resourceType: string, resour
...
+    const DOCKERFILE_PATH_LABEL_KEY = 'dockerfile-path';
+
+    let inputManifestFiles = inputParams.manifests || [];
+    filePathsConfig[MANIFEST_PATHS_KEY] = inputManifestFiles;
+
+    let helmChartPaths = (process.env.HELM_CHART_PATHS && process.env.HELM_CHART_PATHS.split('\n').filter(path => path != "")) || [];
+    filePathsConfig[HELM_CHART_KEY] = helmChartPaths;
+
+    let imageNames = core.getInput('images').split('\n');
+    let imageDockerfilePathMap: any = {};
```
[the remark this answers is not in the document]
Josh-01: Done

src/utilities/utility.ts
```diff
@@ -129,6 +141,62 @@ export function annotateChildPods(kubectl: Kubectl, resourceType: string, resour
...
+    const branchOrTag: string = process.env.GITHUB_REF && process.env.GITHUB_REF.replace('refs/heads/','/').replace('refs/tags/','/');
```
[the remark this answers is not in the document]
Josh-01: Done

src/utilities/utility.ts
```diff
@@ -129,6 +141,62 @@ export function annotateChildPods(kubectl: Kubectl, resourceType: string, resour
...
+            imageConfig = dockerExec.inspectImage(args,true);
+        }
+        catch (ex) {
+            core.warning(`Failed to get dockerfile paths for image ${image.toString()} | ` + ex);
+        }
+
+        if (imageConfig) {
+            imageConfig = JSON.parse(imageConfig)[0];
+            if ((imageConfig.Config) && (imageConfig.Config.Labels) && (imageConfig.Config.Labels[DOCKERFILE_PATH_LABEL_KEY])) {
+                pathValue = imageConfig.Config.Labels[DOCKERFILE_PATH_LABEL_KEY];
```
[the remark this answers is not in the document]
Josh-01: done

src/utilities/utility.ts
```diff
@@ -129,6 +141,62 @@ export function annotateChildPods(kubectl: Kubectl, resourceType: string, resour
...
+            var dockerExec: DockerExec = new DockerExec('docker');
+            dockerExec.pullImage(args,true);
+            imageConfig = dockerExec.inspectImage(args,true);
+        }
+        catch (ex) {
+            core.warning(`Failed to get dockerfile paths for image ${image.toString()} | ` + ex);
+        }
+
+        if (imageConfig) {
+            imageConfig = JSON.parse(imageConfig)[0];
```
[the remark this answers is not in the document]
Josh-01: done

src/docker-object-model.ts
Josh-01: done

src/docker-object-model.ts
Josh-01: Done

src/docker-object-model.ts
Josh-01: Done

src/utilities/utility.ts
```diff
@@ -140,3 +208,11 @@ export function getRandomInt(max: number) {
 export function getCurrentTime(): number {
     return new Date().getTime();
 }
+
+async function checkDockerPath() {
+    var dockerPath = await io.which('docker', false);
+    if (!dockerPath) {
+        throw new Error('Docker is not installed, please use an image with docker');
```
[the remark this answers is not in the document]
Josh-01: Done

src/utilities/utility.ts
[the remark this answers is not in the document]
Josh-01: Done

src/utilities/utility.ts
```diff
@@ -4,6 +4,18 @@ import { IExecSyncResult } from './tool-runner';
 import { Kubectl } from '../kubectl-object-model';
 import { GitHubClient } from '../githubClient';
 import { StatusCodes } from "./httpClient";
+import * as inputParams from "../input-parameters";
+import { DockerExec } from '../docker-object-model';
+import * as io from '@actions/io';
+import * as path from 'path';
+import * as toolCache from '@actions/tool-cache';
+
+
```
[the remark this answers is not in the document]
Josh-01: Done

src/utilities/utility.ts
```diff
@@ -129,6 +141,62 @@ export function annotateChildPods(kubectl: Kubectl, resourceType: string, resour
...
+    let helmChartPaths = (process.env.HELM_CHART_PATHS && process.env.HELM_CHART_PATHS.split('\n').filter(path => path != "")) || [];
+    filePathsConfig[HELM_CHART_KEY] = helmChartPaths;
+
+    let imageNames = core.getInput('images').split('\n');
+    let imageDockerfilePathMap: any = {};
+    const branchOrTag: string = process.env.GITHUB_REF && process.env.GITHUB_REF.replace('refs/heads/','/').replace('refs/tags/','/');
+
+    //Fetching from image label if available
+    for (const image of imageNames) {
+        let args: string[] = [image];
```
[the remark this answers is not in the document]
Josh-01: Done

src/utilities/strategy-helpers/deployment-helper.ts
```diff
@@ -130,15 +130,16 @@ async function checkManifestStability(kubectl: Kubectl, resources: Resource[]):
 
 async function annotateAndLabelResources(files: string[], kubectl: Kubectl, resourceTypes: Resource[], allPods: any) {
     const workflowFilePath = await getWorkflowFilePath(TaskInputParameters.githubToken);
+    const filePathsConfig = await getDeploymentConfig();
```
ajinkya599: `deploymentConfig`

ajinkya599 (approved):
:rocket:

Please do a squash commit while merging the PR:
image

src/utilities/strategy-helpers/deployment-helper.ts
Josh-01: Done

src/constants.ts
koushdey: Yes, this timestamp is later used in two ways in the UI flow. Once as time stamp and the other in readable format. If you are changing it here then changes needs to be done on the UI front as well.

PR #89, opened 2020-12-29
New traceability fields annotation changes - design changes

Josh-01:
Design changes similar to [this PR for v1.3](https://github.com/Azure/k8s-deploy/pull/83).


src/utilities/exec.ts
```diff
@@ -1,33 +0,0 @@
-import * as aexec from '@actions/exec';
```
ajinkya599, line 1: Is this deletion intended?

ajinkya599 (approved):

src/utilities/exec.ts
Josh-01: Yes, this was replaced with ToolRunner

PR #92, opened 2020-12-30
Added kubernetes objects to deployment report data

Ganeshrockz:


src/traceability/azure-traceability-helper.ts
```diff
@@ -77,32 +83,53 @@ async function createDeploymentResource(aksResourceContext: AksResourceContext,
...
-    console.log(`Deployment resource created successfully. Deployment resource object: \n${JSON.stringify(deploymentResource)}`);
-  } catch (error) {
-    console.log(`Some error occured: ${error}`);
+    try {
+      console.log(`Trying to create the deployment resource with payload: \n${JSON.stringify(deploymentPayload)}`);
+      const deploymentResource = await createDeploymentResource(aksResourceContext, deploymentPayload);
+      console.log(`Deployment resource created successfully. Deployment resource object: \n${JSON.stringify(deploymentResource)}`);
+    } catch (error) {
+      console.log(`Some error occured: ${error}`);
+      return Promise.reject(error);
```
ajinkya599, line 97: Let's not make the deployment fail because of traceability.
Change the message to `Some error occured while creating the deployment resource for traceability: '${error}'` ?

src/traceability/azure-traceability-helper.ts
```diff
@@ -77,32 +83,53 @@ async function createDeploymentResource(aksResourceContext: AksResourceContext,
...
 }
 
-export async function addTraceability(): Promise<void> {
+export async function addTraceability(deployedManifestFiles: string[]): Promise<void> {
   const aksResourceContext = getAksResourceContext();
-  const deploymentReport = createDeploymentReport(aksResourceContext);
-  try {
+  if (aksResourceContext !== null) {
+    const deploymentReport = getDeploymentReport(aksResourceContext, deployedManifestFiles);
     const deploymentPayload = getDeploymentPayload(deploymentReport);
```
ajinkya599, line 90: Keep everything in `try-catch` to avoid any failures due to traceability?

src/actions/promote.ts
```diff
@@ -74,6 +75,10 @@ async function promoteBlueGreen(kubectl: Kubectl) {
 
     // checking stability of newly created deployments 
     const deployedManifestFiles = result.newFilePaths;
+
+    // Adding traceability data
+    await AzureTraceabilityHelper.addTraceability(deployedManifestFiles);
```
ajinkya599, line 80: There are some calls to `adjust/route traffic` (in `promote/reject` flows for `canary` and `blue-green`) that are technically modifying the cluster as a part of this deployment. Although the objects of higher interest are the ones related to workloads and they get captured with the current approach.

Perhaps, the ideal hook for capturing traceability seems to be `kubectl.apply` (something like - have a property attached with `kubectl` object to track all deployed manifests. This keeps updating on each call to `apply`. Flush this towards the end of the action execution). Thoughts?

That said, we are currently living with _not capturing_ these objects with `annotations` approach. Maybe it makes sense to continue with that approach.

src/traceability/azure-traceability-helper.ts
```diff
@@ -21,7 +27,7 @@ function getAksResourceContext(): AksResourceContext {
     const rawContent = fs.readFileSync(aksResourceContextPath, 'utf-8');
     return JSON.parse(rawContent);
   } catch (ex) {
-    throw new Error(`An error occured while reading/parsing the contents of the file: ${aksResourceContextPath}. Error: ${ex}`);
+    return null;
```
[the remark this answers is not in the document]
Ganeshrockz: I don't think users will care about this. This is only for our internal reference where we check if the AKSResourcecontext is present or not.

src/actions/promote.ts
Ganeshrockz: Moved stuff to kubectl apply

src/traceability/azure-traceability-helper.ts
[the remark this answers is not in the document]
ajinkya599: If the user comes to us saying that deployment traceability is not working, these logs can help us identify the issue.
We can keep it in debug logs, if we intend to not show the user what we are doing in the background, thus keeping it discoverable only if the user is running in debug mode.

src/traceability/azure-traceability-helper.ts
ajinkya599: Remove `return Promise.reject(error);`?

Ganeshrockz: Are we not failing the run in this case?

src/utilities/files-helper.ts
```diff
@@ -68,6 +69,18 @@ export function writeManifestToFile(inputObjectString: string, kind: string, nam
     return '';
 }
 
+export function getManifestFileContents(filePath: string) {
+    if (filePath &&
+        filePath.length > 0) {
+        try {
+            return yaml.safeLoad(fs.readFileSync(filePath, { encoding: "utf-8" }));
+        } catch (ex) {
+            core.debug('Exception occurred while reading manifest file : ' + filePath + ' . Exception: ' + ex);
```
ajinkya599, line 78: Prefer string interpolation for readability:
`` `Exception occurred while reading manifest file: '${filePath}'. Exception: '${ex}'` ``

src/traceability/azure-traceability-helper.ts
ajinkya599: In my opinion, we should not. We are injecting traceability without user's input. We should not start failing existing workflows.
Maybe, we should add a `warning`. Something to discuss with the PMs? (I believe PMs had this ask to introduce this in the existing action version, and not require user's to start using a newer version).

src/traceability/azure-traceability-helper.ts
```diff
@@ -33,12 +29,10 @@ function getAksResourceContext(): AksResourceContext {
 
 function getDeploymentPayload(deploymentReport: DeploymentReport): any {
   return {
-    // "location": "westus", // Should we set it? If yes, how?
     "properties": {
       "targetResource": {
         "name": deploymentReport.targetResource.properties['name'],
```
ajinkya599, line 34: Let's remove `name` per this spec:
https://github.com/Azure/azure-rest-api-specs-pr/pull/2541

src/kubectl-object-model.ts
```diff
@@ -137,4 +152,31 @@ export class Kubectl {
         if (typeof str === 'string') { return str; }
         return str.join(',');
     }
+
+    private populateDeployedObjects(configurationPaths: string | string[]) {
```
ajinkya599, line 156: Move this to `files-helper`.

src/kubectl-object-model.ts
```diff
@@ -122,6 +133,10 @@ export class Kubectl {
             return this.execute(['delete'].concat(args));
     }
 
+    public getDeployedObjects() {
```
ajinkya599, line 136: [Suggestion]:
- Store `deployedManifestPaths` in `kubectl` object.
- Add method `getManifestObjects(paths: string[]): any[]` in `files-helper`.
- `getDeployedObjects()` calls `getManifestObjects(this.deployedManifestPaths)` and transforms to `IKubeObject`.

This way, the objects are loaded in-memory only when actually needed. Thoughts?

src/kubectl-object-model.ts
Ganeshrockz: This looks like a kubectl related operation rather than a generic file related one. Should we consider moving it to kubectl-utils then? I think the latter looks better.

src/kubectl-object-model.ts
```diff
@@ -122,6 +133,28 @@ export class Kubectl {
...
+                        manifestContent.metadata.name) {
+                            deployedObjects.push({
+                                version: manifestContent.apiVersion,
+                                kind: manifestContent.kind,
+                                name: manifestContent.metadata.name
+                            });
+                        }
+                });
+            }
+        return deployedObjects;
```
ajinkya599, line 155: We should ensure uniqueness of these objects.
In some cases like `canary promote`, `Trafficsplit` is configured twice in the action lifetime. There may be other cases.

ajinkya599 (approved):
🚀

PR #93, opened 2020-12-30
Bug fix for dockerfile path link

Josh-01:
Dockerfile path should be converted from relative path to link in same repo, if relative path is given.


Josh-01:
Wrong commits

PR #96, opened 2020-12-30
Bug fix for dockerfile path for default

Josh-01:
Dockerfile path should be converted to link within current repo when relative path is given by user.

### DIFF
--- a/src/utilities/utility.ts
+++ b/src/utilities/utility.ts
@@ -193,11 +193,11 @@ async function checkDockerPath() {
 function getDockerfilePath(imageConfig: any): string {
     const DOCKERFILE_PATH_LABEL_KEY = 'dockerfile-path';
     const ref: string = process.env.GITHUB_REF && process.env.GITHUB_REF.replace('refs/heads/', '').replace('refs/tags/', '');
-    let pathLabel: string, pathLink: string, pathValue: string = '';
+    let pathLink: string, pathValue: string = '';
     if (imageConfig) {
         if ((imageConfig.Config) && (imageConfig.Config.Labels) && (imageConfig.Config.Labels[DOCKERFILE_PATH_LABEL_KEY])) {
-            pathLabel = imageConfig.Config.Labels[DOCKERFILE_PATH_LABEL_KEY];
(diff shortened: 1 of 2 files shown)


src/utilities/utility.ts
```diff
@@ -193,11 +193,11 @@ async function checkDockerPath() {
 function getDockerfilePath(imageConfig: any): string {
     const DOCKERFILE_PATH_LABEL_KEY = 'dockerfile-path';
     const ref: string = process.env.GITHUB_REF && process.env.GITHUB_REF.replace('refs/heads/', '').replace('refs/tags/', '');
-    let pathLabel: string, pathLink: string, pathValue: string = '';
+    let pathLink: string, pathValue: string = '';
```
ajinkya599, line 196: Move declaration to the scope where it is getting used.

src/utilities/utility.ts
```diff
@@ -193,11 +193,11 @@ async function checkDockerPath() {
...
     const DOCKERFILE_PATH_LABEL_KEY = 'dockerfile-path';
     const ref: string = process.env.GITHUB_REF && process.env.GITHUB_REF.replace('refs/heads/', '').replace('refs/tags/', '');
-    let pathLabel: string, pathLink: string, pathValue: string = '';
+    let pathLink: string, pathValue: string = '';
     if (imageConfig) {
         if ((imageConfig.Config) && (imageConfig.Config.Labels) && (imageConfig.Config.Labels[DOCKERFILE_PATH_LABEL_KEY])) {
-            pathLabel = imageConfig.Config.Labels[DOCKERFILE_PATH_LABEL_KEY];
-            if (pathValue.startsWith('./')) {  //if it is relative filepath convert to link from current repo
+            const pathLabel = imageConfig.Config.Labels[DOCKERFILE_PATH_LABEL_KEY];
+            if (pathLabel.startsWith('./')) {  //if it is relative filepath convert to link from current repo
```
ajinkya599, line 200: Will this check be robust? 🤔 File path may not start with `./`
Should we instead check `if (!isUrl(pathLabel))` ?

Josh-01: Done

ajinkya599 (approved):

PR #108, opened 2021-01-12
Updated deploymentConfig fetch to be called once, populate manifests …

Josh-01:
Made changes such that: 
- 'checkDocker' to be called once while fetching deployment config
- 'deploymentConfig' object created once and used for both annotation and deploymentResource
- Manifests array will be empty when helmcharts present.

### DIFF
--- a/src/actions/promote.ts
+++ b/src/actions/promote.ts
@@ -17,6 +17,7 @@ import { routeBlueGreenIngress, promoteBlueGreenIngress } from '../utilities/str
 import { routeBlueGreenSMI, promoteBlueGreenSMI, cleanupSMI } from '../utilities/strategy-helpers/smi-blue-green-helper';
 import { Kubectl, Resource } from '../kubectl-object-model';
 import * as AzureTraceabilityHelper from "../traceability/azure-traceability-helper";
+import { getDeploymentConfig } from '../utilities/utility';
 
 export async function promote() {
     const kubectl = new Kubectl(await utils.getKubectl(), TaskInputParameters.namespace, true);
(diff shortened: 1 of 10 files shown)


src/utilities/strategy-helpers/deployment-helper.ts
```diff
@@ -55,8 +55,9 @@ export async function deploy(kubectl: Kubectl, manifestFilePaths: string[], depl
         core.debug("Unable to parse pods; Error: " + e);
     }
 
-    await annotateAndLabelResources(deployedManifestFiles, kubectl, resourceTypes, allPods);
-    await AzureTraceabilityHelper.addTraceability(kubectl);
+    const deploymentConfig = await getDeploymentConfig();
+    await annotateAndLabelResources(deployedManifestFiles, kubectl, resourceTypes, allPods, deploymentConfig);
+    await AzureTraceabilityHelper.addTraceability(kubectl, deploymentConfig);
```
ajinkya599, line 60: Make this change in other references of `addTraceability` (I believe `promote` and `reject`)

src/actions/promote.ts
```diff
@@ -25,13 +26,13 @@ export async function promote() {
...
+        // Adding traceability data
+        const deploymentConfig = await getDeploymentConfig();
+        await AzureTraceabilityHelper.addTraceability(kubectl, deploymentConfig);
     } else {
         core.debug('Strategy is not canary or blue-green deployment. Invalid request.');
         throw ('InvalidPromotetActionDeploymentStrategy');
     }
-
-    // Adding traceability data
-    await AzureTraceabilityHelper.addTraceability(kubectl);
```
ajinkya599, line 34: Why move this code?

Josh-01: This will already be called inside the if() block, where deploymentHelper.deploy() is called.
So I moved it to the else if() block to prevent repetition.

ajinkya599: good catch! This was missed earlier.

ajinkya599 (approved):

PR #117, opened 2021-02-28
Added arch detection to kubectl download

### LINKED ISSUE #115: ARM and other arch's support
Hello,

I see in code that kubectl arch is hard-coded to amd64. Is it possible to make it download kubectl depending on agent arch? As I understand `os.arch()` should return correct architecture and if it used, correct kubectl will be downloaded.

gtrubach:
This PR adds the arch detection to kubectl download using [`os.arch()`](https://nodejs.org/api/os.html) call. It won't work on all arch's, but should support amd64, arm and arm64 on Linux (and some more - https://kubernetes.io/docs/setup/release/notes/#client-binaries). If the action is run on non-supported arch by kubectl it won't be able to download and will fail.

Should fix #115.

### DIFF
--- a/src/utilities/kubectl-util.ts
+++ b/src/utilities/kubectl-util.ts
@@ -6,6 +6,7 @@ import * as toolCache from '@actions/tool-cache';
 import * as util from 'util';
 
 import { Kubectl } from '../kubectl-object-model';
+import { StatusCodes } from "./httpClient"
 
 const kubectlToolName = 'kubectl';
 const stableKubectlVersion = 'v1.15.0';
@@ -19,17 +20,25 @@ function getExecutableExtension(): string {
     return '';
 }
 
-export function getkubectlDownloadURL(version: string): string {
+function getKubectlArch(): string {
+    let arch = os.arch();
+    if (arch === 'x64') {
+        return 'amd64';
+    }
+    return arch;
+}
+
+export function getkubectlDownloadURL(version: string, arch: string): string {
     switch (os.type()) {
         case 'Linux':
-            return util.format('https://storage.googleapis.com/kubernetes-release/release/%s/bin/linux/amd64/kubectl', version);
+            return util.format('https://storage.googleapis.com/kubernetes-release/release/%s/bin/linux/%s/kubectl', version, arch);
 
         case 'Darwin':
-            return util.format('https://storage.googleapis.com/kubernetes-release/release/%s/bin/darwin/amd64/kubectl', version);
+            return util.format('https://storage.googleapis.com/kubernetes-release/release/%s/bin/darwin/%s/kubectl', version, arch);
 
         case 'Windows_NT':
         default:
-            return util.format('https://storage.googleapis.com/kubernetes-release/release/%s/bin/windows/amd64/kubectl.exe', version);
+            return util.format('https://storage.googleapis.com/kubernetes-release/release/%s/bin/windows/%s/kubectl.exe', version, arch);
 
     }
 }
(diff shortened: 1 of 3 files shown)


gtrubach:
Integration tests seems to be failing because of ${{ secrets.L2_REPO_TOKEN }} missing.

gtrubach:
To test it I created another branch in my fork kubectl-multiarch-test where I checked-in node_modules. Then I created a repo to test this here https://github.com/gtrubach/k8s-deploy-test which points to the action in that branch and tries to deploy to my local cluster.

The result is here - https://github.com/gtrubach/k8s-deploy-test/runs/1999747697. It uses two self-hosted runners: Windows X64 and Raspberry Pi 2 ARM. Both successful.

Please let me know if anything is needed to be fixed.

src/utilities/kubectl-util.ts
```diff
@@ -19,17 +19,26 @@ function getExecutableExtension(): string {
     return '';
 }
 
+function getKubectlArch(): string {
```
ds-ms, line 22: You need to add a check to validate if `kubectl` can be installed on that `arch` as well.

ds-ms:
Paging @gtrubach to address the comments, so we can take this forward :smile:

src/utilities/kubectl-util.ts
gtrubach: The best way to check if `kubectl` can be installed on particular `arch` is to send a request to the downloadURL and see what is the response. This is because for some versions of `kubectl` the supported arch's are different, also there may darwin arm come in the future.

At the same time the request may fail because the version of `kubectl` is not valid (e.g. 1.30.0 does not exist). Or network error occurred. 

So what I can propose is to check if the exception is an instance of `HTTPError` in downloadKubectl [which is thrown in tool-cache lib](https://github.com/actions/toolkit/blob/master/packages/tool-cache/src/tool-cache.ts#L103) and in case when response is 404 throw an error with more detailed message (like 'The kubectl tool version `<ver>` for arch `<arch>` does not exist.')

What do you think about it?

gtrubach: BTW the default message for HTTPError is `Unexpected HTTP response: ${httpStatusCode}` - https://github.com/actions/toolkit/blob/main/packages/tool-cache/src/tool-cache.ts#L20

ds-ms: Fair enough. 

The `downloadKubectl` doesn't seem to float this error message up, to the logs. 
Could you amend it as well? So that if there's a `404` we can assume it's an arch related issue.

gtrubach: Actually the error is propagated to the log. Here is an example - https://github.com/gtrubach/k8s-deploy-test/runs/2088601164?check_suite_focus=true#step:5:16 (or [raw log](https://pipelines.actions.githubusercontent.com/UAGCXeplpCZe8Boqwsqr09QgMHPfUA9cUS8kBJ3tP8dQqSI059/_apis/pipelines/1/runs/20/signedlogcontent/3?urlExpires=2021-03-11T18%3A47%3A45.1879667Z&urlSigningMethod=HMACV1&urlSignature=1amaEcIOh0vFXGwtrp8tUaaAO0IDj%2BfPzFUUs%2FaaIOg%3D)). With current code if there is a 404 error during kubectl download, the below red message will be shown.

image

The successful example of this action is here - https://github.com/gtrubach/k8s-deploy-test/runs/2088682804?check_suite_focus=true and here for arm linux - https://github.com/gtrubach/k8s-deploy-test/runs/1999747697

src/utilities/kubectl-util.ts
```diff
@@ -51,11 +60,16 @@ export async function getStableKubectlVersion(): Promise<string> {
...
     let cachedToolpath = toolCache.find(kubectlToolName, version);
     let kubectlDownloadPath = '';
+    let arch = getKubectlArch();
     if (!cachedToolpath) {
         try {
-            kubectlDownloadPath = await toolCache.downloadTool(getkubectlDownloadURL(version));
+            kubectlDownloadPath = await toolCache.downloadTool(getkubectlDownloadURL(version, arch));
         } catch (exception) {
-            throw new Error('DownloadKubectlFailed');
+            if (exception instanceof toolCache.HTTPError && exception.httpStatusCode === StatusCodes.NOT_FOUND) {
```
gtrubach, line 68: This is the part which handles 404 cases.

src/utilities/kubectl-util.ts
ds-ms: Cool. This works! 😄

ds-ms:
@gtrubach one last thing left, before I merge. You have to update/add the unit test. You can find the failing test here: https://github.com/Azure/k8s-deploy/pull/117/checks?check_run_id=2088920179

gtrubach:
@ds-ms - Sorry, my bad. I saw the tests failed but was confused with the red error message about token.

I extended that test to run with several test cases and added another test to test that we handle 404 cases.

ds-ms (approved):

Issue #105, opened 2021-01-12
Invalid label workflowFriendlyName


ajinkya599:
It seems like we are adding a string which can have spaces as a label.
<img width="961" alt="image" src="https://user-images.githubusercontent.com/11447401/104274190-4d2d3c00-54c6-11eb-87c9-27aa85f997ab.png">

ds-ms:
@ajinkya599 add your workflow snippet as well for an easy repro

ds-ms:
Closing this issue due to inactivity, feel free to reopen if the problem persists.

ajinkya599:
I think this should be fixed.
cc: @koushdey

koushdey:
Will have a look at it and fix it.

5nafu:
Same issue appears when there are `/` in the name...
```
error: invalid label value: "workflowFriendlyName=CI/CD": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')
```
Is there a possibility to not label the resources?

alexbartsch:
Hej @5nafu,
I also ran into problems with slashes in workflow names, so I made this PR https://github.com/Azure/k8s-deploy/pull/231 to fix this problem.

OliverMKing:
Reopening temporarily. Will get this merged in.

PR #125, opened 2021-03-18
Fixing label values of workflow names with spaces

koushdey:
Fixing issue #105

### DIFF
--- a/lib/utilities/strategy-helpers/deployment-helper.js
+++ b/lib/utilities/strategy-helpers/deployment-helper.js
@@ -154,10 +154,7 @@ function annotateResources(files, kubectl, resourceTypes, allPods, annotationKey
     utility_1.checkForErrors(annotateResults, true);
 }
 function labelResources(files, kubectl, label) {
-    let workflowName = process.env.GITHUB_WORKFLOW;
-    workflowName = workflowName.startsWith('.github/workflows/') ?
-        workflowName.replace(".github/workflows/", "") : workflowName;
-    const labels = [`workflowFriendlyName=${workflowName}`, `workflow=${label}`];
+    const labels = [`workflowFriendlyName=${utility_1.normaliseWorkflowStrLabel(process.env.GITHUB_WORKFLOW)}`, `workflow=${label}`];
     utility_1.checkForErrors([kubectl.labelFiles(files, labels)], true);
 }
 function isCanaryDeploymentStrategy(deploymentStrategy) {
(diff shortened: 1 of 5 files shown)


lib/utilities/strategy-helpers/deployment-helper.js
```diff
@@ -157,7 +157,7 @@ function labelResources(files, kubectl, label) {
     let workflowName = process.env.GITHUB_WORKFLOW;
     workflowName = workflowName.startsWith('.github/workflows/') ?
         workflowName.replace(".github/workflows/", "") : workflowName;
-    const labels = [`workflowFriendlyName=${workflowName}`, `workflow=${label}`];
+    const labels = [`workflowFriendlyName=${workflowName.replace(/ /g, "_")}`, `workflow=${label}`];
```
ds-ms, line 160: What if there were trailing spaces? Like `New   Workflow.yml`; then it would be `New___Workflow.yml` right? Could you fix that to just replace it to `New_Workflow.yml`

src/utilities/strategy-helpers/deployment-helper.ts
```diff
@@ -154,7 +154,7 @@ function labelResources(files: string[], kubectl: Kubectl, label: string) {
     let workflowName = process.env.GITHUB_WORKFLOW;
     workflowName = workflowName.startsWith('.github/workflows/') ?
         workflowName.replace(".github/workflows/", "") : workflowName;
-    const labels = [`workflowFriendlyName=${workflowName}`, `workflow=${label}`];
+    const labels = [`workflowFriendlyName=${workflowName.replace(/ /g, "_")}`, `workflow=${label}`];
```
ajinkya599, line 157: Move the logic to a utility method `normalizeLabel`.
Can any other label have this problem?

lib/utilities/strategy-helpers/deployment-helper.js
ajinkya599: I am kind of okay with multiple underscores. (Retains the original text to some extent and avoids unnecessary complexity).

koushdey: Not sure if we should do this. What should be output if workflow name is "Workflow - Test"? As of now it will be "Workflow_-_Test". Should we normalise such instances for a better workflow friendly name?

src/utilities/strategy-helpers/deployment-helper.ts
koushdey: As of now we are labelling based on custom inputs only in this case. Other labels will not have this issue.

src/utilities/utility.ts
```diff
@@ -130,6 +130,12 @@ export function annotateChildPods(kubectl: Kubectl, resourceType: string, resour
     return commandExecutionResults;
 }
 
+export function normaliseWorkflowStrLabel(workflowStr: string): string {
```
ajinkya599, line 133: Could replace term `workflowStr` by `workflowName` in method and parameter names?

ajinkya599 (approved):

PR #131, opened 2021-04-14
Supporting both comma and new line as delimiters for manifests

Ganeshrockz:
This PR adds support for providing input manifests separated by both commas and newline as delimiter.

The following formats should be supported post this change

```
manifests: |
    a.yml, b.yml, c.yml
```

```
manifests: |
    a.yml
    b.yml, c.yml
```

```
manifests: |
    a.yml
    b.yml
    c.yml
```

```
manifests: |
    a.yml;b.yml;c.yml
```

### DIFF
--- a/__tests__/run.test.ts
+++ b/__tests__/run.test.ts
@@ -294,6 +294,27 @@ test("run() - deploy - Manifests provided by both new line and comma as a delimi
     await expect(action.run()).resolves.not.toThrow();
 });
 
+test("run() - deploy - Manifests provided by both new line and comma and semi-colon as a delimiter", async () => {
+    const kubectlVersion = 'v1.18.0'
+    coreMock.getInput = jest.fn().mockImplementation((name) => {
+        if (name == 'manifests') {
+            return "bg-smi.yml\nbg.yml,deployment.yml;bg.yml";
+        }
+        if (name == 'action') {
+            return 'deploy';
+        }
+        return kubectlVersion;
+    });
+    coreMock.setFailed = jest.fn();
+    toolCacheMock.find = jest.fn().mockReturnValue(undefined);
+    toolCacheMock.downloadTool = jest.fn().mockReturnValue('downloadpath');
+    toolCacheMock.cacheFile = jest.fn().mockReturnValue('cachepath');
+    fileUtility.chmodSync = jest.fn();
+
+    //Invoke and assert
+    await expect(action.run()).resolves.not.toThrow();
+});
+
 test("deployment - deploy() - Invokes with no manifestfiles", async () => {
     const kubeCtl: jest.Mocked<Kubectl> = new Kubectl("") as any;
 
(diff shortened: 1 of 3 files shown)


src/run.ts
```diff
@@ -59,7 +59,13 @@ export async function run() {
         namespace = 'default';
     }
     let action = core.getInput('action');
-    let manifests = manifestsInput.split('\n');
+    let manifests = manifestsInput.split(/[\n,]+/);
+
+    if (manifests.length > 0) {
```
Ganeshrockz, line 64: @shigupt202 I had to add this check explicitly because this looks like a bug. When we provide manifest names with a whitespace it gets added to the file path. Have we heard such cases?

shigupt202: I'm not sure if anyone got into this issue before. But this looks like a valid change.

src/input-parameters.ts
```diff
@@ -5,7 +5,7 @@ import * as core from '@actions/core';
 export let namespace: string = core.getInput('namespace');
 export const containers: string[] = core.getInput('images').split('\n');
 export const imagePullSecrets: string[] = core.getInput('imagepullsecrets').split('\n').filter(secret => secret.trim().length > 0);
-export const manifests = core.getInput('manifests').split('\n');
+export const manifests = core.getInput('manifests').split(/[\n,]+/);
```
shigupt202, line 8: Maybe add the empty string filter that's present for `imagePullSecrets` as well? Also the string[] type.

__tests__/run.test.ts
```diff
@@ -210,6 +210,69 @@ test("run() - deploy - Manifiest not provided", async () => {
     expect(coreMock.setFailed).toBeCalledWith('No manifests supplied to deploy');
 });
 
+test("run() - deploy - Manifests provided by new line delimiter", async () => {
```
shigupt202, line 213: Can we also have one test where only single manifest file is given with no delimiters?

src/input-parameters.ts
```diff
@@ -5,7 +5,7 @@ import * as core from '@actions/core';
 export let namespace: string = core.getInput('namespace');
 export const containers: string[] = core.getInput('images').split('\n');
 export const imagePullSecrets: string[] = core.getInput('imagepullsecrets').split('\n').filter(secret => secret.trim().length > 0);
-export const manifests = core.getInput('manifests').split(/[\n,]+/);
+export const manifests = core.getInput('manifests').split(/[\n,]+/).filter(manifest => manifest.trim().length > 0);
```
shigupt202, line 8: Can you do this at the other reference of `core.getInput('manifests')` as well in run.ts?

shigupt202 (approved):

src/run.ts
```diff
@@ -59,7 +59,13 @@ export async function run() {
         namespace = 'default';
     }
     let action = core.getInput('action');
-    let manifests = manifestsInput.split('\n');
+    let manifests = manifestsInput.split(/[\n,]+/).filter(manifest => manifest.trim().length > 0);
```
ds-ms, line 62: Can't we just refer to the input-parameters one here?

ds-ms (approved):
Add `;` also while you're at it. :) 

LGTM 🚢

PR #141, opened 2021-07-20
Sending a warning instead of throwing error when KUBECONFIG is missing.

### LINKED ISSUE #120: Support running in an in-cluster self-hosted worker without separately setting context
Using a solution like [summerwind/actions-runner-controller](https://github.com/summerwind/actions-runner-controller), it is possible to run a self-hosted worker in a Kubernetes cluster. This self-hosted worker can then use the Kubernetes service account the worker is running under to authenticate to the K8s API to eg. do deployment.

`Azure/k8s-deploy` instead insists on a "context" set by `Azure/k8s-set-context`. It cannot use or refuses to use the Kubernetes native service account authentication (ie. token and connection info mounted at `/var/run/secrets/kubernetes.io/serviceaccount`).

As a rule of thumb, I think if `kubectl` is able to run, `Azure/k8s-deploy` should be able to run. If you feel this is out of scope for `k8s-deploy`, alternatively `k8s-set-context` should be augmented to support the native way.

koushdey:
Fixes #120

### DIFF
--- a/src/run.ts
+++ b/src/run.ts
@@ -42,7 +42,7 @@ async function installKubectl(version: string) {
 
 function checkClusterContext() {
     if (!process.env["KUBECONFIG"]) {
-        throw new Error('Cluster context not set. Use k8ssetcontext action to set cluster context');
+        core.warning('KUBECONFIG env is not explicitly set. Ensure cluster context is set by using k8s-set-context / aks-set-context action.');
     }
 }
 
--- a/lib/run.js
+++ b/lib/run.js
@@ -53,7 +53,7 @@ function installKubectl(version) {
 }
 function checkClusterContext() {
     if (!process.env["KUBECONFIG"]) {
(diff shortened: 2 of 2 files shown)


src/run.ts
```diff
@@ -42,7 +42,7 @@ async function installKubectl(version: string) {
 
 function checkClusterContext() {
     if (!process.env["KUBECONFIG"]) {
-        throw new Error('Cluster context not set. Use k8ssetcontext action to set cluster context');
+        core.warning('Unable to find KUBECONFIG. If Cluster context not set use k8ssetcontext action to set cluster context');
```
ds-ms, line 45: "KUBECONFIG env is not explicitly set. Ensure cluster context is set by using k8s-set-context / aks-set-context action." ?

koushdey: Done

ds-ms (approved):

PR #142, opened 2021-08-02
WIP - update internal function for newer ingress version

### LINKED ISSUE #143: blue-green does not work with latest ingress api version
Repro steps:
- Create a blue-green deployment pipeline configured for ingress strategy. 
- Add an ingress yaml that uses the latest api version 'networking.k8s.io/v1' 

Current behavior:
- During deployment, a green ingress is not created
- During the promotion step, the ingress is not updated
- the green deployment is destroyed as expected but the traffic never changes to the new pods

Expected behavior:
- I expect a new ingress and traffic to change to new pods

xmclark:
Adds conditional logic based on the api version to each section when testing fields on an ingress object. This change fixes #143.


xmclark:
I think this is missing some tests...I need to sort out how to add them in this project. I think I can first produce a failing test, then I can fix it. More WIP.

koushdey:
@sundargs2000 Can you review it?

sundargs2000 (approved):
LGTM 👍

Issue #163, opened 2021-12-20
Cannot apply resources with defined namespaces in manifest


jherrera123:
### Repro steps:

- Obtain a manifest with defined namespaces (e.g. cert-manager, ingress-nginx)
  - https://github.com/jetstack/cert-manager/releases/download/v1.6.1/cert-manager.yaml
  - https://raw.githubusercontent.com/kubernetes/ingress-nginx/controller-v1.1.0/deploy/static/provider/cloud/deploy.yaml
- Add k8s-deploy step that deploys to kubernetes and commit, do not specify a namespece

### Current behaviour:

- Kubernetes deployment fails

```
Error: the namespace from the provided object "kube-system" does not match the namespace "cert-manager". You must pass '--namespace=kube-system' 
```

```
Error: the namespace from the provided object "ingress-nginx" does not match the namespace "default". You must pass '--namespace=ingress-nginx'
```

### Expected behaviour:

- k8s-deploy should not pass namespace=default parameter to kubectl if namespace is not defined in github action

The reasoning is, official instructions for ingress-nginx, cert-manager and others do not need a namespace parameter in kubectl, and if namespace=default is passed, the command fails. I have to split the manifests into multiple manifests per each namespace as a workaround.

OliverMKing:
Hello! We will look into this. Thanks!

davidgamero:
i was able to repro, and just made a PR

davidgamero:
merged the fix. please reopen if issue persists

hansmbakker:
I believe I'm encountering this issue with k8s-deploy@v4 but I cannot reopen this issue.

I'm trying to deploy cert-manager 1.10.1. cert-manager is normally [installed using `kubectl apply -f`](https://cert-manager.io/docs/installation/#default-static-install).
I put the manifest from https://github.com/cert-manager/cert-manager/releases/download/v1.10.1/cert-manager.yaml in my repo and ran the following action:
```
      - name: Deploy cert manager (pre-requisite for actions-runner-controller)
        uses: Azure/k8s-deploy@v4
        with:
          action: deploy
          strategy: basic
          manifests: |
            manifests/deployments/cert-manager-v1.10.1.yml
          force: true
```

Log:
```
Run Azure/k8s-deploy@v4
  with:
    action: deploy
    strategy: basic
    manifests: manifests/deployments/cert-manager-v1.10.1.yml
  
    force: true
    namespace: default
    pull-images: true
    route-method: service
[...]

PR #195, opened 2022-06-14
omit namespace arg for default namespace

davidgamero:
ref #163
- omit namespace arg if namespace == 'default'
- added unit tests back onto main (they still only ran on master)
- bump jest version to fix peer dep conflict

### DIFF
--- a/src/types/kubectl.ts
+++ b/src/types/kubectl.ts
@@ -147,8 +147,9 @@ export class Kubectl {
     if (this.ignoreSSLErrors) {
       args.push("--insecure-skip-tls-verify");
     }
-    args = args.concat(["--namespace", this.namespace]);
-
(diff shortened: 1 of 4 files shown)


src/types/kubectl.ts
```diff
@@ -147,8 +147,9 @@ export class Kubectl {
     if (this.ignoreSSLErrors) {
       args.push("--insecure-skip-tls-verify");
     }
-    args = args.concat(["--namespace", this.namespace]);
-
+    if (this.namespace != "default") {
```
OliverMKing, line 150: we should also check if this.namespace == ""

this.namespace != "default" && this.namespace != ""

OliverMKing (changes requested):

OliverMKing (approved):

PR #215, opened 2022-07-12
Added Traffic split annotations

Vidya2606:


src/strategyHelpers/blueGreen/smiBlueGreenHelper.ts
```diff
@@ -127,7 +127,8 @@ let trafficSplitAPIVersion = ''
 async function createTrafficSplitObject(
    kubectl: Kubectl,
    name: string,
-   nextLabel: string
+   nextLabel: string,
+   annotations: {[key: string]: string} = {}
```
OliverMKing, line 131: how are you letting a user pass this in? You need to add logic to let a user insert it.

We also need to do the same thing for canary.

OliverMKing (changes requested):

src/run.ts
```diff
@@ -18,6 +18,7 @@ export async function run() {
    const action: Action | undefined = parseAction(
       core.getInput('action', {required: true})
    )
+   const annotations: {[key: string]: string} = {}
```
OliverMKing, line 21: you aren't passing in the annotations with core.getInput

action.yml
```diff
@@ -35,6 +35,10 @@ inputs:
       description: 'Traffic split method to be used. Allowed values are pod and smi'
       required: false
       default: 'pod'
+   traffic-split-annotations:
+      description: 'Traffic split annotations in the form of key/value pair can be used'
+      required: false
+      default: 0
```
OliverMKing, line 41: why a default of 0?

src/run.ts
```diff
@@ -38,11 +39,11 @@ export async function run() {
          break
```
OliverMKing, line 39: you likely need to add this to the deploy scenario as well

src/run.ts
OliverMKing: also keep in mind that with core.getInput you will be parsing it as a string. You will have to parse it as an object

OliverMKing (changes requested):

action.yml
```diff
@@ -35,6 +35,9 @@ inputs:
       description: 'Traffic split method to be used. Allowed values are pod and smi'
       required: false
       default: 'pod'
+   traffic-split-annotations:
+      description: 'Traffic split annotations in the form of key/value pair can be used'
```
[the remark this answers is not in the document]
OliverMKing: Also mention that it's relevant only if strategy == canary or blue green

OliverMKing (changes requested):
Looking good! Small doc changes then we should be good.

src/strategyHelpers/deploymentHelper.ts
```diff
@@ -41,7 +41,8 @@ export async function deployManifests(
    files: string[],
    deploymentStrategy: DeploymentStrategy,
    kubectl: Kubectl,
-   trafficSplitMethod: TrafficSplitMethod
+   trafficSplitMethod: TrafficSplitMethod,
+   annotaions: {[key: string]: string} = {}
```
davidgamero, line 45: nit: typo annotaions -> annotations

src/strategyHelpers/blueGreen/smiBlueGreenHelper.ts
```diff
@@ -91,7 +94,11 @@ export async function rejectBlueGreenSMI(
    await cleanupSMI(kubectl, manifestObjects.serviceEntityList)
 }
 
-export async function setupSMI(kubectl: Kubectl, serviceEntityList: any[]) {
+export async function setupSMI(
+   kubectl: Kubectl,
+   serviceEntityList: any[],
+   annotaions: {[key: string]: string} = {}
```
davidgamero, line 100: nit: typo

OliverMKing (approved):

davidgamero (approved):

PR #216, opened 2022-07-27
Add annotation key prefix

Vidya2606:


src/utilities/workflowAnnotationUtils.ts
```diff
@@ -31,19 +24,15 @@ export function getWorkflowAnnotations(
...
 
 export function getWorkflowAnnotationKeyLabel(
    workflowFilePath: string
 ): string {
    const hashKey = require('crypto')
       .createHash('MD5')
-      .update(`${process.env.GITHUB_REPOSITORY}/${workflowFilePath}`)
+      .update(`${ANNOTATION_PREFIX}/${workflowFilePath}`)
       .digest('hex')
    return `githubWorkflow_${hashKey}`
```
OliverMKing, line 37: let's move the hashkey around here. For this function just return `${ANNOTATION_PREFIX}/k8s-deploy`. That means we can remove the workflowFilePath parameter.

OliverMKing (changes requested):

OliverMKing:
Integration tests are failing.

actions.github.com//k8s-deploy is what's being generated. There's an extra /

OliverMKing (approved):

PR #217, opened 2022-07-27
Fixed Blue/Green Strategy Ingress Route-Method Glitch 

### LINKED ISSUE #143: blue-green does not work with latest ingress api version

jaiveerk:
Fixes #143 

Refactoring of this portion of the codebase is soon to follow


src/strategyHelpers/blueGreen/blueGreenHelper.test.ts
```diff
@@ -0,0 +1,80 @@
+import {createWorkloadsWithLabel, deleteWorkloadsAndServicesWithLabel, getManifestObjects, getNewBlueGreenObject, GREEN_LABEL_VALUE, isServiceRouted, NONE_LABEL_VALUE} from './blueGreenHelper'
+import * as bgHelper from './blueGreenHelper'
+import { Kubectl } from '../../types/kubectl'
+import * as fileHelper from '../../utilities/fileUtils'
+
+jest.mock('../../types/kubectl')
+var testObjects
```
OliverMKing, line 7: moving testObjects inside one level to the describe and using let would be cleaner. Var should just be avoided as much as possible

src/strategyHelpers/blueGreen/ingressBlueGreenHelper.ts
```diff
@@ -42,26 +41,27 @@ export async function deployBlueGreenIngress(
...
    //checking if anything to promote
-   if (
-      !validateIngressesState(
-         kubectl,
-         manifestObjects.ingressEntityList,
-         manifestObjects.serviceNameMap
-      )
-   ) {
-      throw 'Ingress not in promote state'
+   var {areValid, invalidIngresses} = validateIngresses(
```
OliverMKing, line 58: why var? const ideally should work or if not use let

src/strategyHelpers/blueGreen/ingressBlueGreenHelper.ts
```diff
@@ -158,34 +159,30 @@ export function validateIngressesState(
...
-            const currentLabel: string =
-               existingIngress?.metadata?.labels[BLUE_GREEN_VERSION_LABEL]
-
-            // if not green label, then wrong configuration
-            if (currentLabel != GREEN_LABEL_VALUE)
-               areIngressesTargetingNewServices = false
-         } else {
-            // no ingress at all, so nothing to promote
-            areIngressesTargetingNewServices = false
+         var isValid = !!existingIngress && existingIngress?.metadata?.labels[BLUE_GREEN_VERSION_LABEL] === GREEN_LABEL_VALUE
```
OliverMKing, line 162: var again

src/strategyHelpers/blueGreen/serviceBlueGreenHelper.ts
```diff
@@ -25,15 +25,15 @@ export async function deployBlueGreenService(
       GREEN_LABEL_VALUE
    )
 
-   // create other non deployment and non service entities
    const newObjectsList = manifestObjects.otherObjects
       .concat(manifestObjects.ingressEntityList)
       .concat(manifestObjects.unroutedServiceEntityList)
    const manifestFiles = fileHelper.writeObjectsToFile(newObjectsList)
-   if (manifestFiles.length > 0) await kubectl.apply(manifestFiles)
+
```
OliverMKing, line 32: removing the length check here will make this throw an error if manifests is an empty array. Is that intended?

src/strategyHelpers/blueGreen/servicesBlueGreenHelper.test.ts
```diff
@@ -0,0 +1,7 @@
+// deployManifests
```
OliverMKing, line 1: remove this file. Probably an intermediate stage with the larger refactor but need to keep things clean

src/strategyHelpers/deploymentHelper.test.ts
```diff
@@ -0,0 +1,7 @@
+// deployManifests
```
OliverMKing, line 1: remove file, keep clean

src/strategyHelpers/deploymentHelper.ts
```diff
@@ -59,16 +59,17 @@ export async function deployManifests(
             core.getInput('route-method', {required: true})
          )
 
-         const {result, newFilePaths} = await Promise.resolve(
+         const {result, newObjectsList} = await Promise.resolve(
             (routeStrategy == RouteStrategy.INGRESS &&
-               deployBlueGreenIngress(kubectl, files)) ||
+               (deployBlueGreenIngress(kubectl, files))) ||
```
OliverMKing, line 64: this has an extra () around deployBlueGreenIngress. I think it makes it harder to read

src/strategyHelpers/deploymentHelper.ts
```diff
@@ -59,16 +59,17 @@ export async function deployManifests(
...
                (routeStrategy == RouteStrategy.SMI &&
                   deployBlueGreenSMI(kubectl, files)) ||
                deployBlueGreenService(kubectl, files)
          )
 
-         checkForErrors([result])
-         return newFilePaths
+
+         checkForErrors([result.result])
```
OliverMKing, line 71: this pattern of result.result is rough. Better names for things will help a ton. You might want to flatten out this object

OliverMKing (changes requested):
Prettier integration test is failing. Prettier will autoformat the code and has a plugin for whatever ide you are using. You'll need to format with prettier before merging

src/strategyHelpers/blueGreen/serviceBlueGreenHelper.ts
jaiveerk: that was not intended, thanks for catching this! not sure how that happened

src/strategyHelpers/blueGreen/ingressBlueGreenHelper.ts
```diff
@@ -227,3 +233,20 @@ export function updateIngressBackend(
...
 }
+
+export function updateIngressBackend(
+   inputObject: any,
+   serviceNameMap: Map<string, string>
+): any {
+   inputObject = JSON.parse(JSON.stringify(inputObject), (key, value) => {
+      if (
+         key.toLowerCase() === BACKEND &&
+         serviceNameMap.has(value.service.name)
```
OliverMKing, line 244: might want to do value?.service?.name

src/strategyHelpers/blueGreen/blueGreenHelper.test.ts
```diff
@@ -0,0 +1,128 @@
...
+         testObjects.deploymentEntityList[0].spec.selector.matchLabels.app
+      ).toBe('nginx')
+   })
+
+   test('correctly makes new blue green object', () => {
+      const modifiedDeployment = getNewBlueGreenObject(
+         testObjects.deploymentEntityList[0],
+         GREEN_LABEL_VALUE
+      )
+      //@ts-ignore
```
OliverMKing, line 43: what's the reason for having the ts-ignores in the expect? that likely should be fixed instead of ignored

OliverMKing (approved):

src/strategyHelpers/blueGreen/blueGreenHelper.test.ts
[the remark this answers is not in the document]
jaiveerk: they're not untyped but when I make jest mock them, they possess methods that the compiler can't see, so it throws a compilation error without these but the code/tests work fine

src/strategyHelpers/blueGreen/blueGreenHelper.test.ts
```diff
@@ -0,0 +1,128 @@
...
+   createWorkloadsWithLabel,
+   deleteWorkloadsAndServicesWithLabel,
+   getManifestObjects,
+   getNewBlueGreenObject,
+   GREEN_LABEL_VALUE,
+   isServiceRouted,
+   NONE_LABEL_VALUE
+} from './blueGreenHelper'
+import * as bgHelper from './blueGreenHelper'
+import {Kubectl} from '../../types/kubectl'
```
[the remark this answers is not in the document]
jaiveerk: already ran prettier, but ran npm format just now and  it didn't change anything

src/strategyHelpers/blueGreen/blueGreenHelper.test.ts
jaiveerk: I'm working on fixing return types in the refactor if that's ok, just because there's kind of a domino effect with changing functions since their usage is pretty scattered

davidgamero (approved):

OliverMKing (approved):

PR #219, opened 2022-07-28
Fixes to the docker with no images

Vidya2606:


src/types/docker.ts
```diff
@@ -8,6 +8,7 @@ export class DockerExec {
    }
 
    public async pull(image: string, args: string[], silent?: boolean) {
+      image = image.trim()
```
OliverMKing, line 11: this won't actually do anything. trim just removes whitespace which wouldn't affect the command

src/utilities/dockerUtils.ts
```diff
@@ -24,6 +24,9 @@ export async function getDeploymentConfig(): Promise<DeploymentConfig> {
    }
 
    const imageNames = core.getInput('images').split('\n') || []
+   if (imageNames.length == 1 && imageNames[0] == '') {
```
OliverMKing, line 27: this doesn't fix the edge case. what if the input is "  "? We would still want an empty array for input.

src/utilities/dockerUtils.ts
```diff
@@ -24,6 +24,9 @@ export async function getDeploymentConfig(): Promise<DeploymentConfig> {
...
+      imageNames[0] = '[]'
```
OliverMKing, line 28: setting it to '[]' doesn't make sense here. we'd want the entire array to be empty if the input is all whitespace

OliverMKing (changes requested):

src/types/docker.ts
```diff
@@ -8,6 +8,8 @@ export class DockerExec {
    }
 
    public async pull(image: string, args: string[], silent?: boolean) {
+      image = image.replace(/^[ ]+/g, '') //Remove whitespace at beginning of the string
```
OliverMKing, line 11: whitespace at the beginning and end of the string doesn't matter. these two additions are unnecessary

src/utilities/dockerUtils.ts
```diff
@@ -23,7 +23,12 @@ export async function getDeploymentConfig(): Promise<DeploymentConfig> {
       )
    }
 
-   const imageNames = core.getInput('images').split('\n') || []
+   let imageNames = core.getInput('images').split('\n') || []
+   imageNames.forEach((element) => {
+      if (element == ' ' || '') {
```
OliverMKing, line 28: we should be checking if the element is whitespace. This is checking if it's ' ' and if it's not it defaults to '' which is false. This will only enter the if statement if the element is ' '. Instead we want to be removing any elements in the imageNames array that is just whitespace

OliverMKing: You can test this out using unit tests or any Node shell. I like to use just the `node` command on the cli to experiment what happens if I'm ever unsure of something's behavior

OliverMKing (changes requested):

src/utilities/dockerUtils.ts
```diff
@@ -23,7 +23,12 @@ export async function getDeploymentConfig(): Promise<DeploymentConfig> {
       )
    }
 
-   const imageNames = core.getInput('images').split('\n') || []
+   let imageNames = core.getInput('images').split('\n') || []
+   imageNames.forEach((element) => {
+      if (element === null || '/^s+$/') {
```
OliverMKing, line 28: This looks like the correct regex but it's just a string. You aren't comparing it to element here.

OliverMKing (changes requested):

OliverMKing:
Going to close this since it's not being actively worked on

PR #228, opened 2022-08-09
Refactor of Blue/Green

jaiveerk:
Refactored blue/green strategy to allow for more readable and flexible code

### DIFF
--- a/src/actions/deploy.ts
+++ b/src/actions/deploy.ts
@@ -6,7 +6,6 @@ import {
    getResources,
    updateManifestFiles
 } from '../utilities/manifestUpdateUtils'
-import {routeBlueGreen} from '../strategyHelpers/blueGreen/blueGreenHelper'
 import {
    annotateAndLabelResources,
    checkManifestStability,
@@ -14,7 +13,6 @@ import {
 } from '../strategyHelpers/deploymentHelper'
 import {DeploymentStrategy} from '../types/deploymentStrategy'
 import {parseTrafficSplitMethod} from '../types/trafficSplitMethod'
-import {parseRouteStrategy} from '../types/routeStrategy'
 
 export async function deploy(
    kubectl: Kubectl,
@@ -51,15 +49,6 @@ export async function deploy(
    )
    await checkManifestStability(kubectl, resourceTypes)
    core.endGroup()
-
-   if (deploymentStrategy == DeploymentStrategy.BLUE_GREEN) {
-      core.startGroup('Routing blue green')
-      const routeStrategy = parseRouteStrategy(
-         core.getInput('route-method', {required: true})
-      )
-      await routeBlueGreen(kubectl, inputManifestFiles, routeStrategy)
-      core.endGroup()
-   }
 
    // print ingresses
    core.startGroup('Printing ingresses')
--- a/src/actions/deploy.ts
+++ b/src/actions/deploy.ts
@@ -18,7 +18,6 @@ export async function deploy(
    kubectl: Kubectl,
    manifestFilePaths: string[],
    deploymentStrategy: DeploymentStrategy,
-   annotations: {[key: string]: string} = {}
 ) {
    // update manifests
    const inputManifestFiles: string[] = updateManifestFiles(manifestFilePaths)
@@ -34,7 +33,6 @@ export async function deploy(
       deploymentStrategy,
       kubectl,
       trafficSplitMethod,
-      annotations
    )
    core.endGroup()
    core.debug('Deployed manifest files: ' + deployedManifestFiles)
--- a/src/actions/promote.ts
+++ b/src/actions/promote.ts
@@ -9,26 +9,27 @@ import {
 import * as models from '../types/kubernetesTypes'
 import * as KubernetesManifestUtility from '../utilities/manifestStabilityUtils'
 import {
+   BlueGreenDeployment,
    BlueGreenManifests,
-   deleteWorkloadsAndServicesWithLabel,
-   deleteWorkloadsWithLabel,
+   deleteGreenObjects,
    getManifestObjects,
    GREEN_LABEL_VALUE,
    NONE_LABEL_VALUE
 } from '../strategyHelpers/blueGreen/blueGreenHelper'
-import {
-   promoteBlueGreenService,
-   routeBlueGreenService
-} from '../strategyHelpers/blueGreen/serviceBlueGreenHelper'
+
 import {
    promoteBlueGreenIngress,
-   routeBlueGreenIngress
-} from '../strategyHelpers/blueGreen/ingressBlueGreenHelper'
+   promoteBlueGreenService,
+   promoteBlueGreenSMI
+} from '../strategyHelpers/blueGreen/promote'
+
 import {
-   cleanupSMI,
-   promoteBlueGreenSMI,
+   routeBlueGreenService,
+   routeBlueGreenIngressUnchanged,
    routeBlueGreenSMI
-} from '../strategyHelpers/blueGreen/smiBlueGreenHelper'
+} from '../strategyHelpers/blueGreen/route'
+
+import {cleanupSMI} from '../strategyHelpers/blueGreen/smiBlueGreenHelper'
 import {Kubectl, Resource} from '../types/kubectl'
 import {DeploymentStrategy} from '../types/deploymentStrategy'
 import {
@@ -120,7 +121,7 @@ async function promoteBlueGreen(
    )
 
    core.startGroup('Deleting old deployment and making new one')
-   let result
+   let result: BlueGreenDeployment
    if (routeStrategy == RouteStrategy.INGRESS) {
       result = await promoteBlueGreenIngress(kubectl, manifestObjects)
    } else if (routeStrategy == RouteStrategy.SMI) {
@@ -132,7 +133,7 @@ async function promoteBlueGreen(
 
    // checking stability of newly created deployments
    core.startGroup('Checking manifest stability')
-   const deployedManifestFiles = result.newFilePaths
+   const deployedManifestFiles = result.deployResult.manifestFiles
    const resources: Resource[] = getResources(
       deployedManifestFiles,
       models.DEPLOYMENT_TYPES.concat([
@@ -146,17 +147,17 @@ async function promoteBlueGreen(
       'Routing to new deployments and deleting old workloads and services'
    )
    if (routeStrategy == RouteStrategy.INGRESS) {
-      await routeBlueGreenIngress(
+      await routeBlueGreenIngressUnchanged(
          kubectl,
-         null,
          manifestObjects.serviceNameMap,
          manifestObjects.ingressEntityList
       )
-      await deleteWorkloadsAndServicesWithLabel(
+
+      await deleteGreenObjects(
          kubectl,
-         GREEN_LABEL_VALUE,
-         manifestObjects.deploymentEntityList,
-         manifestObjects.serviceEntityList
+         manifestObjects.deploymentEntityList.concat(
+            manifestObjects.serviceEntityList
+         )
       )
    } else if (routeStrategy == RouteStrategy.SMI) {
       await routeBlueGreenSMI(
@@ -165,11 +166,7 @@ async function promoteBlueGreen(
          manifestObjects.serviceEntityList,
          annotations
       )
-      await deleteWorkloadsWithLabel(
-         kubectl,
-         GREEN_LABEL_VALUE,
-         manifestObjects.deploymentEntityList
-      )
+      await deleteGreenObjects(kubectl, manifestObjects.deploymentEntityList)
       await cleanupSMI(kubectl, manifestObjects.serviceEntityList)
    } else {
       await routeBlueGreenService(
@@ -177,11 +174,7 @@ async function promoteBlueGreen(
          NONE_LABEL_VALUE,
          manifestObjects.serviceEntityList
       )
-      await deleteWorkloadsWithLabel(
-         kubectl,
-         GREEN_LABEL_VALUE,
-         manifestObjects.deploymentEntityList
-      )
+      await deleteGreenObjects(kubectl, manifestObjects.deploymentEntityList)
    }
    core.endGroup()
 }
--- a/src/actions/promote.ts
+++ b/src/actions/promote.ts
@@ -42,14 +42,13 @@ export async function promote(
    kubectl: Kubectl,
    manifests: string[],
    deploymentStrategy: DeploymentStrategy,
-   annotations: {[key: string]: string} = {}
 ) {
    switch (deploymentStrategy) {
       case DeploymentStrategy.CANARY:
          await promoteCanary(kubectl, manifests)
          break
       case DeploymentStrategy.BLUE_GREEN:
-         await promoteBlueGreen(kubectl, manifests, annotations)
+         await promoteBlueGreen(kubectl, manifests)
          break
       default:
          throw Error('Invalid promote deployment strategy')
@@ -109,7 +108,6 @@ async function promoteCanary(kubectl: Kubectl, manifests: string[]) {
 async function promoteBlueGreen(
    kubectl: Kubectl,
    manifests: string[],
-   annotations: {[key: string]: string} = {}
 ) {
    // update container images and pull secrets
    const inputManifestFiles: string[] = updateManifestFiles(manifests)
@@ -164,7 +162,6 @@ async function promoteBlueGreen(
          kubectl,
          NONE_LABEL_VALUE,
          manifestObjects.serviceEntityList,
-         annotations
       )
       await deleteGreenObjects(kubectl, manifestObjects.deploymentEntityList)
       await cleanupSMI(kubectl, manifestObjects.serviceEntityList)
--- a/src/strategyHelpers/blueGreen/blueGreenHelper.test.ts
+++ b/src/strategyHelpers/blueGreen/blueGreenHelper.test.ts
@@ -1,20 +1,25 @@
 import {
-   createWorkloadsWithLabel,
-   deleteWorkloadsAndServicesWithLabel,
+   deployWithLabel,
+   deleteGreenObjects,
+   fetchResource,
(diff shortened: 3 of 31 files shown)


src/actions/deploy.ts
```diff
@@ -52,15 +48,6 @@ export async function deploy(
    await checkManifestStability(kubectl, resourceTypes)
    core.endGroup()
```
OliverMKing, line 50: this doesn't seem right. If you are removing the routeBlueGreen function call where are you actually routing?

src/actions/promote.ts
```diff
@@ -146,42 +142,33 @@ async function promoteBlueGreen(
...
          manifestObjects.ingressEntityList
       )
-      await deleteWorkloadsAndServicesWithLabel(
+
+      await deleteGreenObjects(
          kubectl,
-         GREEN_LABEL_VALUE,
-         manifestObjects.deploymentEntityList,
-         manifestObjects.serviceEntityList
+         manifestObjects.deploymentEntityList.concat(
```
OliverMKing, line 153: a popular pattern is to swap this to 
```javascript
[].concat(manifestObjects.deploymentEntityList, manifestObjects.serviceEntityList)
```
it's a little more readable about what's happening. It looks like you're calling a method on `manifestObjects.deploymentEntityList` in the original which implies it's more important than `manifestObjects.serviceEntityList`. Same functionality but doing it on an empty list shows the intent that you're merging them a bit better

src/strategyHelpers/blueGreen/blueGreenHelper.test.ts
```diff
@@ -1,20 +1,25 @@
...
 } from './blueGreenHelper'
 import * as bgHelper from './blueGreenHelper'
 import {Kubectl} from '../../types/kubectl'
 import * as fileHelper from '../../utilities/fileUtils'
+import {K8sObject} from '../../types/k8sObject'
+import * as manifestUpdateUtils from '../../utilities/manifestUpdateUtils'
+import {ExecOutput} from '@actions/exec'
 
 jest.mock('../../types/kubectl')
+let testObjects
```
OliverMKing, line 19: move this inside the test scope (the describe). You should probably also reset its value on beforeEach. 

Limit scope as much as possible to prevent future errors

src/strategyHelpers/blueGreen/deploy.ts
```diff
@@ -0,0 +1,139 @@
...
+import {setupSMI} from './smiBlueGreenHelper'
+
+import {routeBlueGreenForDeploy} from './route'
+
+export async function deployBlueGreen(
+   kubectl: Kubectl,
+   files: string[],
+   routeStrategy: RouteStrategy
+): Promise<BlueGreenDeployment> {
+   let blueGreenDeployment: BlueGreenDeployment
```
OliverMKing, line 23: this pattern shows up a lot in the codebase where we create a variable then set it

this is better
```javascript
const blueGreenDeployment = await (async () => {
   switch (routeStrategy) {
      case RouteStrategy.INGRESS:
         return await deployBlueGreenIngress(kubectl, files)
      case RouteStrategy.SMI:
         return await deployBlueGreenSMI(kubectl, files)
      default:
         return deployBlueGreenService(kubectl, files)
   }
})()
```

It's more explicit what the variable will be plus you can use a const for a more functional style to avoid state bugs.

src/strategyHelpers/blueGreen/deploy.ts
```diff
@@ -0,0 +1,139 @@
...
+   }
+
+   core.startGroup('Routing blue green')
+   await routeBlueGreenForDeploy(kubectl, files, routeStrategy)
+   core.endGroup()
+   core.debug(
+      'objects deployed for ' +
+         routeStrategy +
+         ': ' +
+         JSON.stringify(blueGreenDeployment.objects)
```
OliverMKing, line 46: prefer template strings to concatenating with +

``hello ${world}`` is easier to read and faster. Plus you avoid weird javascript bugs with +

OliverMKing (commented):
Got about halfway through reading it. Will read the rest tomorrow. Good stuff

src/actions/deploy.ts
jaiveerk: I moved the call into deploy. All other actions (promote, reject) call the route methods directly (like routeBGSMI or routeBGService) so I didn't think it made sense to have this "routeBlueGreen" method at all, let alone to have it called outside the call to the deployBlueGreen since that function has preset arguments in the calls that it makes that are suited specifically for deployment. Hence the movement to a "routeForDeploy" function

src/strategyHelpers/blueGreen/smiBlueGreenHelper.ts
```diff
@@ -1,163 +1,102 @@
...
+      const newStableService = getStableSMIServiceResource(serviceObject)
+      const newGreenService = getGreenSMIServiceResource(serviceObject)
       newObjectsList.push(newStableService)
       newObjectsList.push(newGreenService)
    })
 
    // create services
-   const manifestFiles = fileHelper.writeObjectsToFile(newObjectsList)
-   await kubectl.apply(manifestFiles)
+   let servicesDeploymentResult: DeployResult = await deployObjects(
```
OliverMKing, line 47: this should be const too. just use const by default unless you absolutely need to use let

src/strategyHelpers/blueGreen/route.ts
```diff
@@ -0,0 +1,147 @@
...
+import * as core from '@actions/core'
+import {TrafficSplitObject} from '../../types/k8sObject'
+
+export async function routeBlueGreenForDeploy(
+   kubectl: Kubectl,
+   inputManifestFiles: string[],
+   routeStrategy: RouteStrategy
+): Promise<BlueGreenDeployment> {
+   // sleep for buffer time
+   const bufferTime: number = parseInt(
```
OliverMKing, line 28: if you're going to move inputs you might as well move this to inputs.ts too

src/strategyHelpers/blueGreen/route.ts
```diff
@@ -0,0 +1,147 @@
...
+export async function routeBlueGreenIngressUnchanged(
+   kubectl: Kubectl,
+   serviceNameMap: Map<string, string>,
+   ingressEntityList: any[]
+): Promise<BlueGreenDeployment> {
+   const objects = ingressEntityList.filter((ingress) =>
+      isIngressRouted(ingress, serviceNameMap)
+   )
+
+   let deployResult = await deployObjects(kubectl, objects)
```
OliverMKing, line 104: use const. not going to mention more but there's other places.

src/strategyHelpers/blueGreen/deploy.ts
```diff
@@ -0,0 +1,139 @@
...
+      kubectl,
+      servicesAndDeployments,
+      GREEN_LABEL_VALUE
+   )
+
+   const otherObjects = manifestObjects.otherObjects.concat(
+      manifestObjects.unroutedServiceEntityList
+   )
+
+   deployObjects(kubectl, otherObjects)
```
OliverMKing, line 102: should this be awaited?

OliverMKing (changes requested):

src/utilities/inputUtils.ts
```diff
@@ -0,0 +1,10 @@
+import * as core from '@actions/core'
```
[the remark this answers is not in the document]
jaiveerk: that makes sense

src/strategyHelpers/deploymentHelper.ts
```diff
@@ -59,17 +61,14 @@ export async function deployManifests(
...
-               deployBlueGreenService(kubectl, files)
+         const {deployResult} = await deployBlueGreen(
+            kubectl,
+            files,
+            routeStrategy
          )
 
-         checkForErrors([workloadDeployment.result])
-         return workloadDeployment.newFilePaths
+         checkForErrors([deployResult.result])
```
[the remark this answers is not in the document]
jaiveerk: I feel like "deployment" is misleading because the resource wrapping the "deployResult" field is called a BlueGreenDeployment but I'll rename the field to be "execResult" since it's more specific/makes clear the fact that the thing being extracted is just a specific result within this larger "result"

src/strategyHelpers/blueGreen/smiBlueGreenHelper.ts
```diff
@@ -1,163 +1,102 @@
...
-         NONE_LABEL_VALUE,
-         annotations
+         svc.metadata.name,
+         NONE_LABEL_VALUE
       )
-   })
+      tsObjects.push(tsObject as TrafficSplitObject)
+   }
+
+   let tsObjectsDeploymentResult = await deployObjects(kubectl, tsObjects)
```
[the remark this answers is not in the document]
jaiveerk: intermediate step leftover from refactor, condensed that into service deployment, good catch 😄

src/strategyHelpers/blueGreen/smiBlueGreenHelper.ts
jaiveerk: will replace lets with const wherever I can

src/strategyHelpers/blueGreen/deploy.ts
```diff
@@ -0,0 +1,139 @@
...
+            return await deployBlueGreenSMI(kubectl, files)
+         default:
+            return await deployBlueGreenService(kubectl, files)
+      }
+   })()
+
+   core.startGroup('Routing blue green')
+   await routeBlueGreenForDeploy(kubectl, files, routeStrategy)
+   core.endGroup()
+   core.debug(
```
OliverMKing, line 37: move inside group

src/strategyHelpers/blueGreen/deploy.ts
```diff
@@ -0,0 +1,139 @@
...
+      kubectl,
+      servicesAndDeployments,
+      GREEN_LABEL_VALUE
+   )
+
+   const otherObjects = [].concat(
+      manifestObjects.otherObjects,
+      manifestObjects.unroutedServiceEntityList
+   )
+
```
OliverMKing, line 99: remove spaces on 99 and 101. Shows that these parts are related

src/strategyHelpers/blueGreen/deploy.ts
```diff
@@ -0,0 +1,139 @@
...
+   // create other non deployment and non service entities
+   const newObjectsList = [].concat(
+      manifestObjects.otherObjects,
+      manifestObjects.ingressEntityList,
+      manifestObjects.unroutedServiceEntityList
+   )
+
+   deployObjects(kubectl, newObjectsList)
```
OliverMKing, line 133: not awaiting this seems wrong.

src/strategyHelpers/blueGreen/promote.ts
```diff
@@ -0,0 +1,84 @@
...
+   manifestObjects
+): Promise<BlueGreenDeployment> {
+   //checking if anything to promote
+   const {areValid, invalidIngresses} = await validateIngresses(
+      kubectl,
+      manifestObjects.ingressEntityList,
+      manifestObjects.serviceNameMap
+   )
+   if (!areValid) {
+      core.debug('are valid was false')
```
OliverMKing, line 25: This debug isn't needed. The error below will show the error state

src/strategyHelpers/blueGreen/reject.ts
```diff
@@ -0,0 +1,89 @@
...
+   BlueGreenManifests,
+   deleteGreenObjects,
+   NONE_LABEL_VALUE
+} from './blueGreenHelper'
+
+import {routeBlueGreenSMI} from './route'
+
+import {cleanupSMI} from './smiBlueGreenHelper'
+
+export interface RejectResult {
```
OliverMKing, line 14: move to types folder

src/strategyHelpers/blueGreen/reject.ts
```diff
@@ -0,0 +1,89 @@
...
+   deleteResult: K8sDeleteObject[]
+   routeResult: BlueGreenDeployment
+}
+
+import {routeBlueGreenIngressUnchanged, routeBlueGreenService} from './route'
```
OliverMKing, line 19: never have imports below other code

src/strategyHelpers/blueGreen/route.ts
```diff
@@ -0,0 +1,146 @@
...
+
+export async function routeBlueGreenForDeploy(
+   kubectl: Kubectl,
+   inputManifestFiles: string[],
+   routeStrategy: RouteStrategy
+): Promise<BlueGreenDeployment> {
+   // sleep for buffer time
+   const bufferTime: number = getBufferTime()
+   if (bufferTime < 0 || bufferTime > 300)
+      throw Error('Version switch buffer must be between 0 and 300 (inclusive)')
```
OliverMKing, line 31: you can move this if statement to inside the input.ts. Would make more sense there since inputs are handled there

src/strategyHelpers/blueGreen/route.ts
```diff
@@ -0,0 +1,146 @@
...
+}
+
+export async function routeBlueGreenIngress(
+   kubectl: Kubectl,
+   serviceNameMap: Map<string, string>,
+   ingressEntityList: any[]
+): Promise<BlueGreenDeployment> {
+   const newObjectsList = []
+
+   ingressEntityList.forEach((inputObject) => {
```
OliverMKing, line 75: mapping would be better here

src/strategyHelpers/blueGreen/route.ts
```diff
@@ -0,0 +1,146 @@
...
+   return {deployResult, objects}
+}
+
+export async function routeBlueGreenSMI(
+   kubectl: Kubectl,
+   nextLabel: string,
+   serviceEntityList: any[]
+): Promise<BlueGreenDeployment> {
+   const tsObjects: TrafficSplitObject[] = []
+   for (const serviceObject of serviceEntityList) {
```
OliverMKing, line 132: mapping better

src/strategyHelpers/blueGreen/smiBlueGreenHelper.ts
```diff
@@ -233,32 +144,39 @@ export async function validateTrafficSplitsState(
...
             element.service === getBlueGreenResourceName(name, STABLE_SUFFIX)
          ) {
-            if (element.weight != MIN_VAL) trafficSplitsInRightState = false
+            trafficSplitsInRightState =
+               trafficSplitsInRightState && element.weight == MIN_VAL
          }
       })
    }
-
+   core.debug('returning ' + String(trafficSplitsInRightState))
```
OliverMKing, line 171: this debug message isn't helpful

A user wouldn't understand what "returning " means + there's no context

OliverMKing (changes requested):

davidgamero (changes requested):

OliverMKing:
Also there's failing tests

src/strategyHelpers/blueGreen/deploy.ts
jaiveerk: I think the spaces preserve readability, is it convention to not have any?

src/strategyHelpers/blueGreen/promote.ts
```diff
@@ -0,0 +1,83 @@
...
+import {validateIngresses} from './ingressBlueGreenHelper'
+import {validateServicesState} from './serviceBlueGreenHelper'
+import {validateTrafficSplitsState} from './smiBlueGreenHelper'
+
+export async function promoteBlueGreenIngress(
+   kubectl: Kubectl,
+   manifestObjects
+): Promise<BlueGreenDeployment> {
+   //checking if anything to promote
+   let {areValid, invalidIngresses} = await validateIngresses(
```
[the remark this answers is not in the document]
jaiveerk: though it's only used in this case, I still don't want to throw from inside the validateIngress function just in case some situation arises where we encounter an ingress out of the correct state but want to just ignore it instead, feel like separating the validation/exception logic keeps the code more modular/keeps responsibility of "validating ingresses" to the validateIngress function and asks it to do nothing more

src/strategyHelpers/blueGreen/promote.ts
```diff
@@ -0,0 +1,83 @@
...
+
+export async function promoteBlueGreenService(
+   kubectl: Kubectl,
+   manifestObjects
+): Promise<BlueGreenDeployment> {
+   // checking if services are in the right state ie. targeting green deployments
+   if (
+      !(await validateServicesState(kubectl, manifestObjects.serviceEntityList))
+   ) {
+      throw new Error('Found services not in promote state')
```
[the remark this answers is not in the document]
jaiveerk: same response as above :)

src/strategyHelpers/blueGreen/deploy.ts
OliverMKing: spacing out everything doesn't make sense and doesn't help readability

if my code is
```
const apples = getApples()
plantApples(apples)

const grape = getGrape()
plantGrape(grape)
```

it's a lot easier to read than

```
const apples = getApples()

plantApples(apples)

const grape = getGrape()

plantGrape(grape)
```

In this particular case I'd remove the spaces because the logic is grouped. You are getting the objects to deploy then deploying.

Some of it is subjective but spacing everything doesn't make code easier to read because it eliminates the benefit of spaces (breaking up logically grouped parts of code)

src/strategyHelpers/blueGreen/promote.ts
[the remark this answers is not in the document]
davidgamero: okay sounds good

davidgamero (dismissed):
lgtm

src/strategyHelpers/blueGreen/deploy.ts
jaiveerk: fair enough, I just thought that none of these three lines are particularly "related" any more than any other collection of lines, plus the fact that prettier makes so many newlines helps separate out these different function calls a bit but it doesn't really matter

OliverMKing (dismissed):

davidgamero (approved):

PR #241, opened 2022-09-22
Fix description about baseline-and-canary-replicas

parroty:
Hi. It's minor one, but this PR intends to clarify the description for `baseline-and-canary-replicas` option.

I believe this option is referenced only in `smiCanaryHelper`.
- https://github.com/Azure/k8s-deploy/blob/main/src/strategyHelpers/canary/smiCanaryHelper.ts#L17


davidgamero:
Thank you for the clarifying contribution!

davidgamero (approved):

OliverMKing (approved):

PR #246, opened 2022-10-12
Add README.md missing parameters

y-luis-rojo:
Added missing parameters `resource-group` and `name`.


README.md
```diff
@@ -110,7 +110,15 @@ Following are the key capabilities of this action:
     <td>Acceptable values: true, false</br>Default value: false.</td>
   </tr>
   <tr>
-    <td>force </br></br>(Optional)</td>
+    <td>resource-group</br></br>(Optional)</td>
+    <td>Name of resource group - Only required if using private cluster</td>
+  </tr>
+  <tr>
+    <td>name</br></br>(Optional)</td>
+    <td>Resource group name - Only required if using private cluster</td>
```
OliverMKing, line 118: The name is actually the name of the private cluster (not the name of the resource group).

I recognize that these values are coming from the action.yml and that needs to be updated too.

OliverMKing (changes requested):
Thanks for this change! This is a needed doc update

OliverMKing (approved):

OliverMKing:
The CLA needs to be agreed to then I'll merge this in. I appreciate these changes!

PR #254, opened 2022-10-26
[WIP] Integration Tests + Labels/Annotations Bug

jaiveerk:


src/strategyHelpers/blueGreen/smiBlueGreenHelper.ts
```diff
@@ -178,13 +178,13 @@ export async function cleanupSMI(
...
 
    serviceEntityList.forEach((serviceObject) => {
-      deleteList.push({
-         name: getBlueGreenResourceName(
-            serviceObject.metadata.name,
-            TRAFFIC_SPLIT_OBJECT_NAME_SUFFIX
-         ),
-         kind: TRAFFIC_SPLIT_OBJECT
-      })
+      // deleteList.push({
```
OliverMKing, line 181: better to remove than just uncomment here. people won't remember the context and it will just be confusing

.github/workflows/run-integration-tests-bluegreen-ingress.yml
```diff
@@ -0,0 +1,192 @@
...
+   workflow_dispatch:
+
+jobs:
+   run-integration-test:
+      name: Run Minikube Integration Tests
+      runs-on: ubuntu-latest
+      env:
+         KUBECONFIG: /home/runner/.kube/config
+      steps:
+         - uses: actions/checkout@v2
```
OliverMKing, line 22: Lots of these have the same basic pattern of building the action + installing a few other dependencies. Would be a good chance to use [reusable workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows) so we only have to maintain this once.

.github/workflows/run-integration-tests-basic.yml
```diff
@@ -0,0 +1,77 @@
...
+      branches:
+         - master
+         - main
+         - 'releases/*'
+   workflow_dispatch:
+
+jobs:
+   run-integration-test:
+      name: Run Minikube Integration Tests
+      runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
jaiveerk: I don't believe we can since different strategies call different combinations of parameters, and at different stages in each test, based on the parameters we're calling each test with, we call the python test script at different frequencies and with different parameters.

I might be missing something though, what were you thinking?

.github/workflows/run-integration-tests-bluegreen-ingress.yml
jaiveerk: agreed, especially for the linkerd setup. Do you think it makes sense to leave that for a future PR?

OliverMKing: IMO it's easier to manage if it's in one so might be worth it to do now.

OliverMKing:
Tests are all currently failing

.github/workflows/run-integration-tests-basic.yml
```diff
@@ -0,0 +1,76 @@
...
+         matrix:
+            os: [ubuntu-latest, macos-12, windows-latest]
+   
+      name: Run Minikube Integration Tests
+      runs-on: ${{ matrix.os }}
+      env:
+         KUBECONFIG: /home/runner/.kube/config
+         NAMESPACE: test-${{ github.run_id }}
+      steps:
+         - uses: actions/checkout@v2
```
OliverMKing, line 27: v3 is latest and needed. (it will complain because v2 is a node12 action)

PR #255, opened 2022-10-27
Added Integration Tests, Resolved Bugs With Annotations

jaiveerk:


.github/workflows/run-integration-tests-basic.yml
```diff
@@ -0,0 +1,77 @@
...
+         KUBECONFIG: /home/runner/.kube/config
+      steps:
+         - uses: actions/checkout@v2
+
+         - name: Install dependencies
+           run: |
+              rm -rf node_modules/
+              npm install
+
+         - name: Install ncc
```
OliverMKing, line 29: Should be no reason to have this be a separate step. It's installed with `npm install` from the previous step.

.github/workflows/run-integration-tests-basic.yml
```diff
@@ -0,0 +1,77 @@
...
+           run: npm i -g @vercel/ncc
+         - name: Build
+           run: ncc build src/run.ts -o lib
```
OliverMKing, line 32: Let's run `npm run build` instead which does this exact thing. That abstracts the exact tool away which will make it easier to update the build tooling if we ever need to change it.

.github/workflows/run-integration-tests-basic.yml
```diff
@@ -0,0 +1,77 @@
...
+
+         - name: Set name of ns
```
OliverMKing, line 34: Even if we don't switch to the matrix yet it makes sense to make this the env variable. It's just more flexible that way and a more common practice. Setting it in an env variable is the right way to do it vs setting it in a step is more of a hack.

.github/workflows/run-integration-tests-basic.yml
```diff
@@ -0,0 +1,77 @@
...
+           run: echo "::set-output name=name::$(echo `date +%Y%m%d%H%M%S`)"
+           shell: bash
+           id: ns
+
+         - uses: Azure/setup-kubectl@v1
```
OliverMKing, line 39: Use v3 of setup kubectl. v1 is very outdated

src/actions/promote.ts
```diff
@@ -77,22 +80,27 @@ async function promoteCanary(kubectl: Kubectl, manifests: string[]) {
...
-      await SMICanaryDeploymentHelper.redirectTrafficToStableDeployment(
-         kubectl,
-         manifests
-      )
+      const stableRedirectManifests =
+         await SMICanaryDeploymentHelper.redirectTrafficToStableDeployment(
+            kubectl,
+            manifests
+         )
+      promoteResult.manifestFiles.push(...stableRedirectManifests)
```
OliverMKing, line 98: feels weird that we are pushing to a property of promote result.

Can we rename promote result to what they are used for later?

filesToAnnotate maybe? and instead of being type DeployResult it could just be a list of files. That would help with readability

src/strategyHelpers/blueGreen/deploy.ts
```diff
@@ -56,21 +65,35 @@ export async function deployBlueGreenSMI(
...
       manifestObjects.deploymentEntityList,
       GREEN_LABEL_VALUE
    )
-   return {
-      deployResult: blueGreenDeployment.deployResult,
-      objects: [].concat(blueGreenDeployment.objects, newObjectsList)
-   }
+
+   blueGreenDeployment.objects.push(...newObjectsList)
+   blueGreenDeployment.objects.push(...smiAndSvcDeployment.objects)
```
OliverMKing, line 87: not very clear from just the names what the difference between objects and deploy result is. Is there a good name we could change objects to?

src/types/kubectl.ts
```diff
@@ -98,6 +98,14 @@ export class Kubectl {
       return await this.execute(args)
    }
 
+   public async annotateFile(
```
OliverMKing, line 101: is this needed vs annotate files? AnnotateFiles works with a single file passed in

OliverMKing (changes requested):

.github/workflows/run-integration-tests-basic.yml
jaiveerk: ok cool was wondering why we had it, it's just leftover from the previous version so I'll get rid of that

src/actions/promote.ts
jaiveerk: yeah agreed, doesn't make sense to push to the result/to bundle it that way for no reason. will update

src/strategyHelpers/blueGreen/deploy.ts
jaiveerk: the deploy result stores JSON objects of the k8s objects it deploys as well as the manifest files that those objects come from, so figured that would all be contained within one result coming back from when we deploy something. I could change it to k8sobjects if you think that's better but feel like that's implied?

src/types/kubectl.ts
jaiveerk: didn't realize it would work with a single one, but yeah if that's the case then no there shouldn't be any need

src/strategyHelpers/blueGreen/deploy.ts
OliverMKing: Seems fine for now!

.github/workflows/run-integration-tests-bluegreen-ingress.yml
```diff
@@ -0,0 +1,184 @@
...
+              python test/integration/k8s-deploy-delete.py 'Deployment' 'nginx-deployment' ${{ env.NAMESPACE }}
+
+         - if: ${{ always() }}
+           name: Delete created namespace
+           run: kubectl delete ns ${{ env.NAMESPACE }}
+
+         - if: ${{ always() }}
+           name: Posting result back to PR
+           run: |
+              if [ '${{ steps.job-type.outputs.type }}' == 'pr' ]; then ruby postStatus.rb ${{github.event.client_payload.repository}} ${{github.event.client_payload.commit}} ${{secrets.L2_REPO_TOKEN}} ${{job.status}} ${{github.run_id}} ${{matrix.os}} false ${{ secrets.L2_REPO_USER }}; fi
```
OliverMKing, line 183: What's the point of this? If the job fails that seems like enough feedback

OliverMKing (approved):

.github/workflows/run-integration-tests-bluegreen-ingress.yml
jaiveerk: it was left over from the last iteration of integration tests, I can remove it if you think there's no point in keeping it

OliverMKing: Yeah it doesn't do anything. Worth it to remove stuff.

OliverMKing (approved):

ShantingLiu (approved):

richardsimko:
I believe a loop introduced in this PR is causing #264. What was the reason for looping over all files when each command labels all files at once?

PR #257, opened 2022-11-17
Fixed Private Cluster Namespace Bug

### LINKED ISSUE #256: Bug: deployment.apps not found even though the deployment exists
### What happened?

The action uses a deployment file and it successfully deploys the application. But when it goes to check the status of deployment (Checking manifest stability), it throws "Error from server (NotFound): deployments.apps "test-deployment" not found.

I had used this action before and was working well. Not sure how but suddenly I started receiving this error.
Below is the configuration,
```YAML
- name: Deploy to AKS
  uses: Azure/k8s-deploy@v4.6
  with:
    resource-group: ${{ env.AKS_RESOURCE_GROUP_NAME }}
    name: ${{ env.AKS_CLUSTER_NAME }}
    action: deploy
    namespace: 'default'
    manifests: .k8s/test-deployment.yml
    images: ${{ env.REGISTRY_NAME }}.azurecr.io/test-repo:${{ github.sha }}
    private-cluster: true
    annotate-namespace: false
```

### Version

### Runner

GitHub hosted runner

### Relevant log output

image

jaiveerk:
Fixes #256


.github/workflows/run-integration-tests-private.yml
```diff
@@ -0,0 +1,90 @@
...
+           name: Install Kubectl
+
+         - name: Create private AKS cluster and set context
+           run: |
+              set +x
+              # create cluster
+              az group create --location eastus --name ${{ env.NAMESPACE }}
+              az aks create --name ${{ env.NAMESPACE }} --resource-group ${{ env.NAMESPACE }} --enable-private-cluster --generate-ssh-keys
+              echo "getting assignee objectId..."
+              export ASSIGNEE_OBJECT_ID="$(az ad sp show --id ${{ secrets.AZURE_CLIENT_ID }} --only-show-errors| jq -r '.objectId')"
```
OliverMKing, line 49: What does the assignee object id actually do here? I can't tell

.github/workflows/run-integration-tests-private.yml
```diff
@@ -0,0 +1,90 @@
...
+              echo "::add-mask::${ASSIGNEE_OBJECT_ID}"
+              if [ $ASSIGNEE_OBJECT_ID = "null" ]
+              then
+                export ASSIGNEE_OBJECT_ID="$(az ad sp show --id ${{ secrets.AZURE_CLIENT_ID }} --only-show-errors| jq -r '.id')"
+                echo "::add-mask::${ASSIGNEE_OBJECT_ID}"
+              fi
+              echo "assigning role to service principal..."
```
OliverMKing, line 56: are we assigning role to sp? It looks like we are just getting the kubeconfig on the next line

OliverMKing (commented):
Rest looks good to me :). 

Just want to confirm making the default namespace will be a future change? It doesn't look like it's here

src/types/privatekubectl.ts
```diff
@@ -93,23 +106,32 @@ export class PrivateKubectl extends Kubectl {
...
-      let start = strToParse.indexOf('-filename')
-      let offset = 7
-
-      if (start == -1) {
-         start = strToParse.indexOf('-f')
-
-         if (start == -1) {
-            return ''
+      const fileNames: string[] = []
+      const argv = require('minimist')(strToParse.split(' '))
```
davidgamero, line 110: we should probably move this import to the top for clarity ex `minimist =  require('minimist')` or ideally an es6 import if it supports that `import * as minimist from 'minimist'`

jaiveerk:
> 

It's in there! Might have gotten lost among other files: https://github.com/Azure/k8s-deploy/pull/257/files#diff-1243c5424efaaa19bd8e813c5e6f6da46316e63761421b3e5f5c8ced9a36e6b6

.github/workflows/run-integration-tests-private.yml
jaiveerk: good catch! left over from a different approach.

.github/workflows/run-integration-tests-private.yml
jaiveerk: yep yep just left over from a different approach I was trying earlier

src/types/privatekubectl.ts
```diff
@@ -93,23 +106,32 @@ export class PrivateKubectl extends Kubectl {
...
+      const fArg = 'f'
+      const filenameArg = 'filename'
+
+      if (argv[fArg]) {
```
davidgamero, line 114: can we abstract this logic into a method that runs on both the args instead of having two copies of the same logic right after each other?

jaiveerk:
> Rest looks good to me :).
> 
> Just want to confirm making the default namespace will be a future change? It doesn't look like it's here

It's in there! Might have gotten lost among other files: https://github.com/Azure/k8s-deploy/pull/257/files#diff-1243c5424efaaa19bd8e813c5e6f6da46316e63761421b3e5f5c8ced9a36e6b6

src/types/privatekubectl.ts
jaiveerk: makes sense! changing it to es6 to keep it clean

davidgamero (commented):
just a couple small changes requested

davidgamero (approved):
clean LGTM

OliverMKing (approved):

jaiveerk:
Moved to [PR 258](https://github.com/Azure/k8s-deploy/pull/258) due to unsigned commit, closing this PR

PR #259, opened 2022-11-22
Fixing Ubuntu Runner Issue

jaiveerk:
Replacing setup-minikube action with another one that is compatible with new set of latest ubuntu runners in integration tests


OliverMKing:
It would be better to change the action rather than the runner. We need to be sure that our actions work with the latest runner.

OliverMKing (changes requested):
^

OliverMKing (approved):

elenavillamil (approved):

PR #265, opened 2022-11-29
Fixed Resources Being Annotated Multiple Times

### LINKED ISSUE #264: Bug: Labeling resources multiple times
### What happened?

The latest version (4.7) is trying to label all resources multiple times. I believe the issue is with this loop which was introduced in #255: https://github.com/Azure/k8s-deploy/blob/b9a9965750c9d436e6136fcf42a3566508673fd1/src/strategyHelpers/deploymentHelper.ts#L259-L266

Each call labels all resources yet it makes one call for each file. This means that for a large deployment (We have about 20 manifests being deployed at once) it takes unreasonably long to deploy as it's repeating the same action over and over again. For our deployment it takes about 3 minutes to run on 4.7 and 17 seconds on 4.6.

I'm happy to fix the issue but I want to get some context as to why the loop was added in the first place as I don't want to break some other part of the workflow.

### Version

### Runner

github hosted

### Relevant log output

On 4.6 there is only a single line of:

```
  /usr/local/bin/kubectl label -f [all manifests being deployed] workflowFriendlyName=deploy workflow=actions.github.com-k8s-deploy --overwrite --insecure-skip-tls-verify
```

But on 4.7 it's repeating that line once for each file.

jaiveerk:
Fixes #264

### DIFF
--- a/src/strategyHelpers/deploymentHelper.ts
+++ b/src/strategyHelpers/deploymentHelper.ts
@@ -258,7 +258,7 @@ async function labelResources(
    const labelResults = []
    for (const file of files) {
       try {
-         const labelResult = await kubectl.labelFiles(files, labels)
+         const labelResult = await kubectl.labelFiles(file, labels)
          labelResults.push(labelResult)
       } catch (e) {
          core.warning(`failed to annotate resource: ${e}`)


OliverMKing (approved):

bfoley13 (approved):
lgtm!

PR #270, opened 2022-12-07
Check for error from Azure when using the private-cluster feature

oddsund:
Move the error check for Azure earlier, so that a well defined error is thrown on error instead of a JSONSyntax error.

The issue is that when Azure returns an error, like when there's an issue with the access to the principal used. When this happens, the stdout field will be an empty string, and the error message will be set to a non-json string.

### DIFF
--- a/src/types/privatekubectl.ts
+++ b/src/types/privatekubectl.ts
@@ -75,13 +75,17 @@ export class PrivateKubectl extends Kubectl {
             runOutput
          )}`
       )
+
+      if (runOutput.exitCode !== 0) {
+         throw Error(
+            `Call to private cluster failed. Command: '${kubectlCmd}', errormessage: ${runOutput.stderr}`
+         )
+      }
+
       const runObj: {logs: string; exitCode: number} = JSON.parse(
          runOutput.stdout
       )
       if (!silent) core.info(runObj.logs)
-      if (runOutput.exitCode !== 0 && runObj.exitCode !== 0) {
-         throw Error(`failed private cluster Kubectl command: ${kubectlCmd}`)
-      }
 
       return {
          exitCode: runObj.exitCode,
--- a/src/types/privatekubectl.ts
+++ b/src/types/privatekubectl.ts
@@ -86,6 +86,9 @@ export class PrivateKubectl extends Kubectl {
          runOutput.stdout
       )
       if (!silent) core.info(runObj.logs)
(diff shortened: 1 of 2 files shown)


oddsund:

oddsund:

jaiveerk (dismissed):

src/types/privatekubectl.ts
```diff
@@ -75,13 +75,17 @@ export class PrivateKubectl extends Kubectl {
             runOutput
          )}`
       )
+
+      if (runOutput.exitCode !== 0) {
```
OliverMKing, line 79: This part looks good to me and makes sense so we don't fail on json parsing like you said.

src/types/privatekubectl.ts
```diff
@@ -75,13 +75,17 @@ export class PrivateKubectl extends Kubectl {
...
+         throw Error(
+            `Call to private cluster failed. Command: '${kubectlCmd}', errormessage: ${runOutput.stderr}`
+         )
+      }
+
       const runObj: {logs: string; exitCode: number} = JSON.parse(
          runOutput.stdout
       )
       if (!silent) core.info(runObj.logs)
-      if (runOutput.exitCode !== 0 && runObj.exitCode !== 0) {
```
OliverMKing, line 82: We need to keep this part here
```TypeScript
if (runOutput.exitcode !== 0) {
         throw Error(`failed private cluster Kubectl command: ${kubectlCmd}`)
}
```

Without this we won't behave exactly like the [overridden method](https://github.com/Azure/k8s-deploy/blob/e3e5e845b7c25d360b46bd6115c23e2f52a5ed9a/src/types/kubectl.ts#L173). `getExecOutput` will [throw if the exit code is none zero](https://github.com/actions/toolkit/blob/main/packages/exec/src/interfaces.ts#L28) so we need to replicate that behavior here.

OliverMKing (changes requested):

OliverMKing (approved):

src/types/privatekubectl.ts
OliverMKing: Looks good to me! I'll squash merge it when we merge into main. Thanks!

davidgamero (approved):
Lgtm

OliverMKing:
Going to squash and merge this. Will kick off a release sometime soon.

PR #276, opened 2023-01-09
Fixes multiple namespaces bug

### LINKED ISSUE #163: Cannot apply resources with defined namespaces in manifest

OliverMKing:
Closes #163 

Lets users deploy resources in multiple namespaces elegantly

### DIFF
--- a/src/actions/deploy.ts
+++ b/src/actions/deploy.ts
@@ -56,7 +56,9 @@ export async function deploy(
    for (const ingressResource of ingressResources) {
       await kubectl.getResource(
          KubernetesConstants.DiscoveryAndLoadBalancerResource.INGRESS,
-         ingressResource.name
+         ingressResource.name,
+         false,
+         ingressResource.namespace
       )
    }
    core.endGroup()
--- a/src/strategyHelpers/canary/canaryHelper.ts
+++ b/src/strategyHelpers/canary/canaryHelper.ts
@@ -217,7 +217,7 @@ async function cleanUpCanary(
       for (const inputObject of parsedYaml) {
          const name = inputObject.metadata.name
          const kind = inputObject.kind
-         const ns = inputObject?.metadata?.namespace
+         const namespace = inputObject?.metadata?.namespace
 
          if (
             isDeploymentEntity(kind) ||
@@ -227,8 +227,8 @@ async function cleanUpCanary(
             const canaryObjectName = getCanaryResourceName(name)
             const baselineObjectName = getBaselineResourceName(name)
 
-            await deleteObject(kind, canaryObjectName, ns)
-            await deleteObject(kind, baselineObjectName, ns)
+            await deleteObject(kind, canaryObjectName, namespace)
+            await deleteObject(kind, baselineObjectName, namespace)
          }
       }
    }
--- a/src/strategyHelpers/canary/canaryHelper.ts
+++ b/src/strategyHelpers/canary/canaryHelper.ts
@@ -217,7 +217,7 @@ async function cleanUpCanary(
       for (const inputObject of parsedYaml) {
          const name = inputObject.metadata.name
          const kind = inputObject.kind
-         const namespace = inputObject?.metadata?.namespace
+         const namespace: string | undefined = inputObject?.metadata?.namespace
 
          if (
             isDeploymentEntity(kind) ||
--- a/src/strategyHelpers/canary/canaryHelper.ts
+++ b/src/strategyHelpers/canary/canaryHelper.ts
@@ -198,10 +198,10 @@ async function cleanUpCanary(
    const deleteObject = async function (
       kind: string,
       name: string,
-      ns: string
+      namespace: string | undefined
    ) {
       try {
-         const result = await kubectl.delete([kind, name])
+         const result = await kubectl.delete([kind, name], namespace)
          checkForErrors([result])
       } catch (ex) {
          // Ignore failures of delete if it doesn't exist
--- a/src/strategyHelpers/canary/canaryHelper.ts
+++ b/src/strategyHelpers/canary/canaryHelper.ts
@@ -195,7 +195,11 @@ async function cleanUpCanary(
    files: string[],
    includeServices: boolean
 ): Promise<string[]> {
-   const deleteObject = async function (kind, name) {
+   const deleteObject = async function (
+      kind: string,
+      name: string,
+      ns: string
+   ) {
       try {
          const result = await kubectl.delete([kind, name])
          checkForErrors([result])
@@ -213,6 +217,7 @@ async function cleanUpCanary(
       for (const inputObject of parsedYaml) {
          const name = inputObject.metadata.name
          const kind = inputObject.kind
+         const ns = inputObject?.metadata?.namespace
 
          if (
             isDeploymentEntity(kind) ||
@@ -222,8 +227,8 @@ async function cleanUpCanary(
             const canaryObjectName = getCanaryResourceName(name)
             const baselineObjectName = getBaselineResourceName(name)
 
-            await deleteObject(kind, canaryObjectName)
-            await deleteObject(kind, baselineObjectName)
+            await deleteObject(kind, canaryObjectName, ns)
+            await deleteObject(kind, baselineObjectName, ns)
          }
       }
    }
--- a/src/strategyHelpers/deploymentHelper.ts
+++ b/src/strategyHelpers/deploymentHelper.ts
@@ -218,7 +218,12 @@ async function annotateResources(
    )
    if (annotateNamespace) {
       annotateResults.push(
-         await kubectl.annotate('namespace', namespace, annotationKeyValStr)
+         await kubectl.annotate(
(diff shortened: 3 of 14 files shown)


src/utilities/manifestStabilityUtils.ts
```diff
@@ -97,37 +112,47 @@ export async function checkPodStatus(
...
+         resource.namespace
+      )
    }
 }
 
-async function getPodStatus(kubectl: Kubectl, podName: string) {
-   const podResult = await kubectl.getResource('pod', podName)
+async function getPodStatus(kubectl: Kubectl, pod: Resource) {
+   const podResult = await kubectl.getResource(
+      'pod',
```
davidgamero, line 151: nit: magic string

src/utilities/manifestStabilityUtils.ts
```diff
@@ -85,8 +100,8 @@ export async function checkPodStatus(
    for (let i = 0; i < iterations; i++) {
       await sleep(sleepTimeout)
 
-      core.debug(`Polling for pod status: ${podName}`)
-      podStatus = await getPodStatus(kubectl, podName)
+      core.debug(`Polling for pod status: ${resource.name}`)
```
davidgamero, line 103: if this method is just for pods, then does this name change make it harder to understand the code?

src/actions/deploy.ts
```diff
@@ -56,7 +56,9 @@ export async function deploy(
    for (const ingressResource of ingressResources) {
       await kubectl.getResource(
          KubernetesConstants.DiscoveryAndLoadBalancerResource.INGRESS,
-         ingressResource.name
+         ingressResource.name,
```
jaiveerk, line 59: not sure if this is out of scope for this PR, but wonder if we can standardize K8s objects to have namespace in the same field (ie metadata.namespace)? feel like having all objects look the same would make it easier to work with them

src/utilities/kubectlUtils.ts
```diff
@@ -30,7 +30,12 @@ export async function getLastSuccessfulRunSha(
    annotationKey: string
 ): Promise<string> {
    try {
-      const result = await kubectl.getResource('namespace', namespaceName)
+      const result = await kubectl.getResource(
+         'namespace',
```
jaiveerk, line 34: nit: can we make this resource type a constant value

src/strategyHelpers/deploymentHelper.ts
```diff
@@ -218,7 +218,12 @@ async function annotateResources(
    )
    if (annotateNamespace) {
       annotateResults.push(
-         await kubectl.annotate('namespace', namespace, annotationKeyValStr)
+         await kubectl.annotate(
+            'namespace',
```
jaiveerk, line 222: again, not sure how reasonable this is but would love to make this value constant

src/types/kubectl.ts
```diff
@@ -3,11 +3,11 @@ import {createInlineArray} from '../utilities/arrayUtils'
 import * as core from '@actions/core'
 import * as toolCache from '@actions/tool-cache'
 import * as io from '@actions/io'
-import {exec} from 'child_process'
 
 export interface Resource {
    name: string
    type: string
+   namespace?: string
```
jaiveerk, line 10: nit: in addition to maybe moving name and namespace into metadata, can we move Resource to k8sobjects.ts?

src/types/kubectl.ts
```diff
@@ -118,82 +128,90 @@ export class Kubectl {
...
-   protected getExecuteFlags(): string[] {
+   protected getFlags(namespaceOverride?: string): string[] {
       const flags = []
       if (this.ignoreSSLErrors) {
          flags.push('--insecure-skip-tls-verify')
       }
-      if (this.namespace) {
-         flags.push('--namespace', this.namespace)
+
+      const ns = namespaceOverride || this.namespace
```
jaiveerk, line 212: love this, super elegant solution 🔥

src/utilities/kubectlUtils.ts
```diff
@@ -53,12 +58,13 @@ export async function annotateChildPods(
    kubectl: Kubectl,
    resourceType: string,
    resourceName: string,
+   namespace: string | undefined,
```
jaiveerk, line 61: is there a reason why this isn't just `namespace?: string`

src/utilities/manifestStabilityUtils.ts
jaiveerk: I think we should keep the arg as the resource type but would agree that we could make the param name `pod` or something

jaiveerk (changes requested):
looks good for the most part!

can we also modify/add to our integration tests to ensure that multi-namespaced deployments work as intended?

src/actions/deploy.ts
OliverMKing: There's a difference between the actual k8s object and what our app passes around / parses. Standardizing is good but the Resource type is something our code transforms the actual k8s object into so it will have a different model. The Resource type is for our application and our application doesn't need to know about the concept of metadata vs something else.

If we make Resource type the same as a k8s object then a developer would have to know what's useful for this application vs what's not. That's a lot harder than just using a simple object that's parsed from a more complicated one.

src/utilities/kubectlUtils.ts
OliverMKing: optional params need to be after other ones.

src/actions/deploy.ts
jaiveerk: makes sense!

jaiveerk (approved):

bfoley13 (approved):

src/strategyHelpers/canary/canaryHelper.ts
```diff
@@ -213,6 +217,7 @@ async function cleanUpCanary(
       for (const inputObject of parsedYaml) {
          const name = inputObject.metadata.name
          const kind = inputObject.kind
+         const ns = inputObject?.metadata?.namespace
```
Tatsinnit, line 220: 💡 Thought to share:  with this `optional changing` what happens with the scenario, we supply `undefined` in the end? or we might want to add some error handlers in place?

src/strategyHelpers/canary/canaryHelper.ts
```diff
@@ -222,8 +227,8 @@ async function cleanUpCanary(
             const canaryObjectName = getCanaryResourceName(name)
             const baselineObjectName = getBaselineResourceName(name)
 
-            await deleteObject(kind, canaryObjectName)
-            await deleteObject(kind, baselineObjectName)
+            await deleteObject(kind, canaryObjectName, ns)
```
Tatsinnit, line 230: 💡 Nit: various places we use `namespaces, for clarity might be `ns` can be renamed to `namespaces`

src/types/kubectl.ts
```diff
@@ -56,16 +56,24 @@ export class Kubectl {
    public async describe(
       resourceType: string,
       resourceName: string,
-      silent: boolean = false
+      silent: boolean = false,
+      namespace?: string
```
Tatsinnit, line 60: 💡thought to share:  Nullish with `?` but type is passed as `string` not `string | null`

OliverMKing: I was curious about that too. Do you know what the best practice is?

I opted for just `string` instead of `string | undefined` because that's what they do in the [typescript official docs](https://www.typescriptlang.org/docs/handbook/2/functions.html).

Tatsinnit: ☕️🥷 indeed! ++ , thinking out loud - For anything unintuitive it’s better to mention type upfront. Totally with you, thinking about the function contract for types.

Given you have thought this far, I will go with whatever decision we choose and we have it archived here anyways.

src/strategyHelpers/canary/canaryHelper.ts
OliverMKing: Added some additional typing to make it more explicit.

PR #283, opened 2023-04-02
Fix “Service” route-method of the Blue-Green strategy with some manifest files

otetard:
With the current implementation of the “Service” route-method of the Blue-Green strategy, if the manifest file contains `Service` object _before_ `Deployment` ones, then, `Service` isn’t updated with the right labels selector.

<details><summary>Sample affected manifest file</summary>

```yaml
apiVersion: v1
kind: Service
metadata:
   name: nginx-service
spec:
   selector:
      app: nginx
   ports:
      - protocol: TCP
        port: 80
        targetPort: 80
---
apiVersion: apps/v1
kind: Deployment
metadata:
   name: nginx-deployment
   labels:
      app: nginx
spec:
   replicas: 1
   selector:
      matchLabels:
         app: nginx
   template:
      metadata:
         labels:
            app: nginx
      spec:
         containers:
            - name: nginx
              image: nginx
              ports:
                 - containerPort: 80
```
</details>

To fix this issue, update `getManifestObjects()` function in the `src/strategyHelpers/blueGreen/blueGreenHelper.ts` file to first extract all objects from the manifest file, and _then_ identify “routed” service entities. This way, even if the `Deployment` object is declared after the “routed” `Service`, Blue-Green strategy will be able to _not_ update the `Service` until the routing switch (done in `routeBlueGreenForDeploy()`, in `src/strategyHelpers/blueGreen/route.ts`).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,8 +22,8 @@
             "@types/jest": "^26.0.0",
             "@types/js-yaml": "^3.12.7",
             "@types/node": "^12.20.41",
+            "@vercel/ncc": "^0.36.1",
             "jest": "^26.0.0",
-            "ncc": "^0.3.6",
             "prettier": "^2.7.1",
             "ts-jest": "^26.0.0",
             "typescript": "3.9.5"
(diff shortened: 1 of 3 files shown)


otetard:

OliverMKing (approved):
This is a good catch. Thank you for the PR. We will get it merged in and released.

jaiveerk (approved):

aamgayle (approved):

PR #289, opened 2023-06-14
Add scaledjob support.

x-projs:


davidgamero (approved):
lgtm

jaiveerk:
can you add UTs for this change?

src/types/kubernetesTypes.ts
```diff
@@ -32,6 +32,7 @@ export const WORKLOAD_TYPES: string[] = [
    'replicaset',
    'daemonset',
    'pod',
+   'scaledjob',
```
jaiveerk, line 35: can you add a UT for this change in kubernetesTypes.test.ts?

jaiveerk (changes requested):

PR #297, opened 2023-10-30
release workflow

davidgamero:
add new release workflow to upgrade our release process to better fit github actions best practices

### DIFF
--- a/.github/workflows/release-pr.yml
+++ b/.github/workflows/release-pr.yml
@@ -1,14 +1,16 @@
-name: Create release PR
+name: release Project
 
 on:
+   push:
+      branches:
+         - main
+      paths:
+         - CHANGELOG.md
    workflow_dispatch:
-      inputs:
-         release:
-            description: 'Define release version (ex: v1, v2, v3)'
-            required: true
 
 jobs:
-   release-pr:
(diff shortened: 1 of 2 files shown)


jaiveerk (approved):

.github/workflows/release-pr.yml
```diff
@@ -1,14 +1,15 @@
...
-      inputs:
-         release:
-            description: 'Define release version (ex: v1, v2, v3)'
-            required: true
 
 jobs:
-   release-pr:
-      uses: OliverMKing/javascript-release-workflow/.github/workflows/release-pr.yml@main
+   release:
+      uses: OliverMKing/action-release-workflows/.github/workflows/release_js_project.yaml@main
```
OliverMKing, line 13: We should probably move the action-release workflows into an Azure repo. Also unsure about using @main instead of a sha.

davidgamero: done

OliverMKing (approved):

bramdehart:
@davidgamero, @jaiveerk , @OliverMKing  unable to retrieve the new release for the Github action. Added an issue: https://github.com/Azure/k8s-deploy/issues/300

PR #299, opened 2023-10-31
bump release workflow sha

davidgamero:
the current workflow uses a short hash which apparently isn't supported


jaiveerk (approved):

OliverMKing (approved):

OliverMKing:
Have you tested this in a fork? We should be testing this in a fork instead of against the main repo so we aren't clogging up the history

PR #302, opened 2023-11-06
add installing ncc to build

davidgamero:
use a local bin link just for ncc to get a bin link prior to calling with npx, as it isn't working correctly on our action runners

using just `npx ncc` isn't working since the install/build step doesn't add bin links, and we have to manually add one for ncc which we use for building

link to a working version on my fork https://github.com/davidgamero/k8s-deploy/actions/runs/6751019098/job/18354429862

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
    "author": "Deepak Sattiraju",
    "license": "MIT",
    "scripts": {
-      "build": "npx ncc build src/run.ts -o lib",
+      "build": "npm i ncc && ncc build src/run.ts -o lib",
       "test": "jest",
       "coverage": "jest --coverage=true",
       "format": "prettier --write .",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -4,7 +4,7 @@
    "author": "Deepak Sattiraju",
    "license": "MIT",
    "scripts": {
-      "build": "npx ncc build src/run.ts -o lib",
+      "build": "npm i ncc && ncc build src/run.ts -o lib",
```
jaiveerk: does this not need to be -g?

also unrelated do we want to change the author to AKS Atlanta

jaiveerk (changes requested):

package.json
davidgamero: yep, but i meant to still use npx, just use npm i without the --no-bin-links so that npx can find it

davidgamero: that way we dont have to install globally and can use the one from the package.json still

jaiveerk (approved):

bfoley13 (approved):

PR #303, opened 2023-11-08
fix: ensure imageNames are not empty strings

Foxboron:
In Typescript/Javascript an empty string split on newline is going to produce an array with an empty string.

    => "".split('\n')
    [""]

This causes the action to produce a warning, unless `pull-images` is set to false.

    Failed to get dockerfile path for image : Error: The process '/usr/bin/docker' failed with exit code 1

Filtering the list to remove any zero-length strings from the array solves this issue.

### DIFF
--- a/src/utilities/dockerUtils.ts
+++ b/src/utilities/dockerUtils.ts
@@ -23,7 +23,11 @@ export async function getDeploymentConfig(): Promise<DeploymentConfig> {
       )
    }
 
-   const imageNames = core.getInput('images').split('\n') || []
+   const imageNames =
+      core
+         .getInput('images')
+         .split('\n')
+         .filter((image) => image.length > 0) || []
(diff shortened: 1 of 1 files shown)


Foxboron:
Note the previously signed CLA is not valid for this code contribution.

davidgamero (approved):
lgtm thanks for the contribution!

davidgamero:
@microsoft-github-policy-service rerun

davidgamero:
do you mean because of the minimum line/file [requirements](https://github.com/microsoft/.github/blob/14c530c4637d42a67e9b57c04b9231f78c10e6b7/policies/cla.yml#L12C1-L14C23)?

Foxboron:
Ah, if that is the portion that triggers then that is fine. I thought I had singed it for my personal contributors at some point and can't tell form the check why it was green.

Feel free to merge, thanks :)

PR #316, opened 2024-03-12
create v5 node20 release

davidgamero:
create the new v5 release with node20 updates and the required workflow version bumps

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -10,6 +10,8 @@ jobs:
    CodeQL-Build:
       # CodeQL runs on ubuntu-latest and windows-latest
       runs-on: ubuntu-latest
+      permissions:
+         contents: read
 
       steps:
          - name: Checkout repository
@@ -44,3 +46,4 @@ jobs:
 
          - name: Perform CodeQL Analysis
            uses: github/codeql-action/analyze@v3
+         
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -21,7 +21,7 @@ jobs:
 
          # Initializes the CodeQL tools for scanning.
          - name: Initialize CodeQL
-           uses: github/codeql-action/init@v2
+           uses: github/codeql-action/init@v3
            # Override language selection by uncommenting this and choosing your languages
            # with:
            #   languages: go, javascript, csharp, python, cpp, java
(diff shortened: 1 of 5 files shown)


jaiveerk (approved):

.github/workflows/codeql.yml
```diff
@@ -10,26 +10,29 @@ jobs:
    CodeQL-Build:
       # CodeQL runs on ubuntu-latest and windows-latest
       runs-on: ubuntu-latest
+      permissions:
+         contents: read
+         security-events: write
 
       steps:
          - name: Checkout repository
-           uses: actions/checkout@v3
```
hsubramanianaks, line 16: Are we supposed to ping to sha instead of versions ex: https://github.com/Azure/Bridge-To-Kubernetes/blob/main/.github/workflows/codeql-analysis.yml#L49

CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 # Changelog
 
-## [v4.10.0] - 2023-10-30
```
hsubramanianaks, line 3: also change log PR would always have only changes related to release, not other bump's ? like github version bump's. if we are following different strategy then fine. Just calling it out.

hsubramanianaks (approved):
left some comments as suggestions, please take a look. Thank you.

.github/workflows/codeql.yml
jaiveerk: we're moving over to semantic versioning/pinning major versions since that's what the community overwhelmingly asked for (even though it's not technically best practice)

OliverMKing: I agree with @hsubramanianaks. Our workflows should use sha, especially when pointing towards actions not owned by us.

What we agreed on wasn't to use the tag in our workflows, just to expose the major version tag when we release our workflows (but recommend using the sha in the action docs). This matches what github owned actions do. That is not related to this.

OliverMKing (changes requested):

.github/workflows/codeql.yml
davidgamero: yes good points i've update with hashes

CHANGELOG.md
davidgamero: sure i'll add the other PRs here

CristianAtMaersk:
The new version and tags are missing still.
The release workflow failed https://github.com/Azure/k8s-deploy/actions/runs/8350336943

PR #320, opened 2024-03-31
Make namespace input optional

florianmutter:
This allows to deploy manifests in multiple different namespaces. `annotate-namespace` cannot be used if no namespace is given.


florianmutter:

davidgamero:
thank you for this contribution!

can we add an integration test to cover this?

adding a new directory in the [integration manifests](https://github.com/Azure/k8s-deploy/tree/main/test/integration/manifests) with two example manifests with different namespaces, and then adding a new workflow to cover that scenario would do it

it'd be similar to the workflow for the [basic integration test](https://github.com/Azure/k8s-deploy/blob/main/.github/workflows/run-integration-tests-basic.yml)

Issue #350, opened 2024-11-18
Feature Request: Allow setting timeout in watching rollout status


davidgamero:
### Feature request

enable setting a timeout for watching a rollout via the `kubectl rollout status` --timeout flag
https://kubernetes.io/docs/reference/kubectl/generated/kubectl_rollout/kubectl_rollout_status/

PR #353, opened 2024-11-19
adding timeout to the rollout status

Vidya2606:
Adding timeout to the rollout status - https://github.com/Azure/k8s-deploy/issues/350

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -80,6 +80,10 @@ inputs:
    skip-tls-verify:
       description: True if the insecure-skip-tls-verify option should be used. Input should be 'true' or 'false'.
       default: false
+   timeout:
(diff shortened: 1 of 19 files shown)


davidgamero:
this looks good.
let's add the input to `action.yaml` so it can be configured by users
this means it will be retrieved in `run.ts` via `core.getInput` and then passed down to the functions that use it.

also, let's default to a longer default input in action.yaml like 10m

davidgamero (approved):
lgtm

benjaminbob21:
I will be closing this PR as there is another PR #425 that fixes this same issue, thanks for the initial implementation @Vidya2606.

PR #398, opened 2025-04-16
case-insensitive resource type

davidgamero:
- allow case-insensitive matching of resource types
- centralize core.getInput retrieval to run.ts for `resourceType`
- add tests for case-insensitive resource type and throw on unknown resource type


bfoley13 (approved):
Looks good

Issue #243, opened 2022-10-08
Feature Request: CroJobs


juzerali:
### Feature request

Requesting support for CronJobs. Currently image substitution in CronJobs does not work.

```yml
  deploy:
    name: Deploy
    needs: [ build, publish ]
    runs-on: ubuntu-latest
    steps:
      - name: Set the Kubernetes context
        uses: azure/k8s-set-context@v2
        with:
          method: service-account
          k8s-url: https://my-kube-domain
          k8s-secret: ${{ secrets.KUBERNETES_SECRET }}

      - name: Checkout source code
        uses: actions/checkout@v3

      - name: Deploy to the Kubernetes cluster
        uses: azure/k8s-deploy@v1
        with:
          namespace: default
          manifests: |
            kube/cronjob.yml
          images: |
            asia-south1-docker.pkg.dev/project/repo/image:${{ github.sha }}
```

Expected: CronJob is updated with the new image as per `${{ github.sha }}`
Actual: Image name that is hardcoded, gets updated

OliverMKing:
I'm going to look into this!

OliverMKing:
You can temporarily use https://github.com/Azure/k8s-artifact-substitute this action to swap images prior to running deploy. It should correctly handle this case. Simply take the output of this action and use that as your manifests input in deploy.

PR #433, opened 2025-07-08
Feature/cronjob image substitution

maddox-214:
🚀 Add image substitution support for CronJobs (Addresses #243 )
This PR enhances the image substitution logic to support CronJobs by updating container image references in structured YAML objects. It replaces fragile string-based substitution with robust object-level mutation, ensuring compatibility across all standard workload types including CronJob, Deployment, and initContainers.


src/utilities/manifestUpdateUtils.ts
```diff
@@ -81,68 +81,89 @@ function updateContainerImagesInManifestFiles(
...
+      let updatedObjects: K8sObject[] = []
+      inputObjects.forEach((obj) => {
+         if (isWorkloadEntity(obj.kind)) {
+            containers.forEach((container: string) => {
+               let [imageName] = container.split(':')
+               if (imageName.indexOf('@') > 0) {
+                  imageName = imageName.split('@')[0]
+               }
+               updateImagesInK8sObject(obj, imageName, container)
+            })
```
Tatsinnit, line 96: 💡 How about something like this and will it work, key takes are below.

* readFileSync now uses 'utf8' encoding directly. `map` replaces manual `forEach + push`.
* includes('@') is more expressive than indexOf('@') > 0.

```*.ts
const fileContents = fs.readFileSync(filePath, 'utf8');
const inputObjects = yaml.loadAll(fileContents) as K8sObject[];

const updatedObjects = inputObjects.map((obj) => {
  if (!isWorkloadEntity(obj.kind)) {
    return obj;
  }

  containers.forEach((container: string) => {
    let [imageName] = container.split(':');
    if (imageName.includes('@')) {
      imageName = imageName.split('@')[0];
    }
    updateImagesInK8sObject(obj, imageName, container);
  });

  return obj;
});

```

src/utilities/manifestUpdateUtils.ts
```diff
@@ -81,68 +81,89 @@ function updateContainerImagesInManifestFiles(
...
+      obj?.kind?.toLowerCase() === KubernetesWorkload.CRON_JOB &&
+      obj?.spec?.jobTemplate?.spec?.template?.spec?.initContainers
+   ) {
+      updateImageInContainerArray(
+         obj.spec.jobTemplate.spec.template.spec.initContainers,
+         imageName,
+         newImage
+      )
+   }
+}
```
Tatsinnit, line 149: 💡 How about

```*.ts
function updateImagesInK8sObject(
   obj: any,
   imageName: string,
   newImage: string
) {
   const isCronJob = obj?.kind?.toLowerCase() === KubernetesWorkload.CRON_JOB;

   const containerPaths = [
      // Regular workload
      obj?.spec?.template?.spec,
      // CronJob workload
      isCronJob ? obj?.spec?.jobTemplate?.spec?.template?.spec : null,
   ].filter(Boolean); // Remove any undefined/null entries

   for (const path of containerPaths) {
      if (path?.containers) {
         updateImageInContainerArray(path.containers, imageName, newImage);
      }
      if (path?.initContainers) {
         updateImageInContainerArray(path.initContainers, imageName, newImage);
      }
   }
}


```

src/utilities/manifestUpdateUtils.ts
maddox-214: Yes this looks cleaner, will implement. Thank you!

src/utilities/manifestUpdateUtils.ts
```diff
@@ -79,70 +79,71 @@ function updateContainerImagesInManifestFiles(
    filePaths: string[],
```
bosesuneha, line 79: Can you add some unit tests for this?

maddox-214: Will do, thanks

src/utilities/manifestUpdateUtils.test.ts
```diff
@@ -25,4 +25,83 @@ describe('manifestUpdateUtils', () => {
...
+                  }
+               }
+            }
+         }
+         manifestUpdateUtils['updateImagesInK8sObject'](obj, 'init', 'init:new')
+         expect(
+            obj.spec.jobTemplate.spec.template.spec.initContainers[0].image
+         ).toBe('init:new')
+      })
+   })
```
bosesuneha, line 106: this looks good, can we add a test for updateContainerImagesInManifestFiles to validate that the yaml parsing works as expected

Issue #288, opened 2023-06-12
Feature Request: more spesific rollout error


Elias-Chairi:
### Feature request

When checking rollout status I expect to see the actual status for deployment/pod-status like `ImagePullBackOff` instead of `Error: Rollout status error`.

PR #440, opened 2025-07-15
Feature Request: Add Enhanced Deployment Error Reporting and Logging

ejaifeobuks:
This PR enhances deployment diagnostics by providing more specific and actionable error reporting during rollout failures. Key improvements include:

Improved Manifest Stability: Adds detailed error reporting and logging to help identify deployment issues more precisely.
Aggregated Rollout Errors: Collects and throws a comprehensive error message summarizing all rollout failures.
Container Diagnostics: Introduces getContainerErrors to extract container status for better pod-level troubleshooting.
Verbose Failure Logging: Captures detailed kubectl describe outputs for rollout status, pod, and service checks to aid debugging.
Additionally, this PR includes unit tests to verify:

Proper aggregation of detailed error messages.
Accurate logging of describe outputs during failure scenarios.
This PR will close issue #288


src/utilities/manifestStabilityUtils.test.ts
```diff
@@ -3,6 +3,12 @@ import {Kubectl} from '../types/kubectl'
 import {ResourceTypeFleet, ResourceTypeManagedCluster} from '../actions/deploy'
 import {ExecOutput} from '@actions/exec'
 import {exitCode, stdout} from 'process'
+import * as core from '@actions/core'
+import {sleep} from './timeUtils'
+
+jest.mock('./timeUtils', () => ({
```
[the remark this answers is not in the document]
ejaifeobuks: Hey David!
It doesn't belong to a specific test. The aim is for it to serve all tests for functions that have a wait time to execute. Also, since it's at the module level, Jest.restoreAllMocks() can't clear it

Tatsinnit (approved):
Thank you so much for this, all changes look fairly comprehensive, so it will be super great to get regular eyes for review like @bosesuneha or @davidgamero , I noticed something which you could simplify but its for inspiration only: https://github.com/Azure/k8s-deploy/pull/440/files#r2213055224 

Thank you once again for this PR. ❤️

src/utilities/manifestStabilityUtils.ts
```diff
@@ -182,6 +244,42 @@ function isPodReady(podStatus: any): boolean {
    return allContainersAreReady
 }
 
+export function getContainerErrors(podStatus: any): string {
+   const errors: string[] = []
+
+   if (podStatus.containerStatuses) {
+      podStatus.containerStatuses.forEach((container) => {
+         if (!container.ready) {
```
davidgamero, line 252: can this by refactored to use an early return instead? that would reduce the branch depth here

ex:
 ```ts
  if (ready) return;
```

also destructing the container would make this easier to read:
```ts
podStatus.containerStatuses.forEach(({ name, ready, state }) => {
`

Comment (ejaifeobuks) [reply src/utilities/manifestStabilityUtils.ts]: Thanks for this. I have put out a new commit that addresses these.

File: src/utilities/manifestStabilityUtils.ts [hunk 3]
```diff
@@ -182,6 +244,42 @@ function isPodReady(podStatus: any): boolean {
    return allContainersAreReady
 }
 
+export function getContainerErrors(podStatus: any): string {
```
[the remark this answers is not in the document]
ejaifeobuks: Thank you for the suggestion, Tats. It was very helpful in streamlining the function and reducing its length. I’ve implemented your feedback, and a new commit with these changes is now available.

src/utilities/manifestStabilityUtils.ts
```diff
@@ -182,6 +244,33 @@ function isPodReady(podStatus: any): boolean {
...
+   const errors: string[] = []
+   const collectErrors = (containers: any[], label: string) => {
+      containers?.forEach(({name, ready, state}) => {
+         if (!ready) {
```
davidgamero, line 251: as this condition bypasses the remaining execution, i would recommend an early return here
` if (ready) return`

or 

```ts
if (ready){
  return
}
```

src/utilities/manifestStabilityUtils.ts
```diff
@@ -38,20 +46,37 @@ export async function checkManifestStability(
...
       }
 
-      if (resource.type == KubernetesConstants.KubernetesWorkload.POD) {
+      if (
+         resource.type.toLowerCase() ===
+         KubernetesConstants.KubernetesWorkload.POD.toLowerCase()
+      ) {
          try {
-            await checkPodStatus(kubectl, resource)
+            await exports.checkPodStatus(kubectl, resource)
```
davidgamero, line 79: why is this exports.checkPodStatus instead of just checkPodStatus?

ejaifeobuks: The reason is to make it testable. The **checkPodStatus** function is called by **checkManifestStability** within the same module. When functions call each other internally within the same module, Jest spies cannot intercept those calls by default. By exporting **checkPodStatus** and using **exports.checkPodStatus()** instead of a direct call, the function goes through the module's export object, which allows Jest spies to successfully intercept and mock the function call during testing.

ejaifeobuks: Unless I am misunderstanding you, this approach won't work either because we're already importing the entire '**manifestStabilityUtils'** module in the tests. If we don't want to export c**heckPodStatus**, our only alternatives are to either not test the checkPodStatus function in isolation, or move it to a separate file. The fundamental problem remains that since **checkManifestStability** is in the same file as **checkPodStatus,** any call from **checkManifestStability** to **checkPodStatus** will be internal and without the export the **Jest.Spy** will not be able to intercept this call

PR #528, opened 2026-05-30
fix: confine manifest paths to workspace and harden URL fetcher

gambtho:
## Summary

- Confines the `manifests:` input to `GITHUB_WORKSPACE`. Paths that resolve outside the workspace (`../` traversal, absolute paths outside, or symlinks pointing outside) now fail the action with a clear error instead of being read and copied. Files copied into `RUNNER_TEMP` now use a basename-only destination with a counter+timestamp uniquifier, so the copy step cannot write outside `RUNNER_TEMP` even if validation were bypassed.
- Fixes error handling in `writeYamlFromURLToFile`: drops a misleading `async` on the `https.get` callback that swallowed throws, adds `error` listeners on the response and file-writer streams, returns after rejecting HTTP errors so the success branch no longer runs, drains the response on early reject, and guards the verification body so a null YAML document (e.g. `---`) cannot hang the promise.
- Emits `core.warning` when `GITHUB_WORKSPACE` is unset (the runner always sets it; unset signals a misconfigured environment, not a reason to silently skip the containment check).

### Background — path traversal

`moveFileToTmpDir()` previously built its destination as `path.join(tempDirectory, originalFilepath)`. `path.join` normalizes `../` rather than blocking it, so a `manifests` input of `../../.config/myapp/secrets.yaml` would (a) read a YAML file from outside the workspace and (b) write a copy to a path outside `RUNNER_TEMP`. Directory inputs made this stronger because `recurisveManifestGetter` recursively enumerated all `.yaml`/`.yml` files under the traversed directory.

The fix:
- Adds `assertPathWithinWorkspace`, which uses `fs.realpathSync` on both the workspace and the input (defeating symlink escapes) and compares with `path.relative` (handling exact `..`, `../foo`, absolute paths, and `..bar.yaml`-style filenames correctly).
- Applies it at the entry to `getFilesFromDirectoriesAndURLs` so URLs bypass the check while disk inputs are validated before `lstatSync`/`readdirSync`.
- Re-validates every leaf in `recurisveManifestGetter` so a symlink inside an accepted directory that points outside the workspace is still rejected.
- Rewrites `moveFileToTmpDir` to a basename-only destination under `RUNNER_TEMP` with `\${base}_\${getCurrentTime()}_\${moveCounter++}\${ext}`, mirroring (and improving on) the safer pattern already present in `getNewTempManifestFileName`.

This is a **breaking change** for workflows that rely on manifest paths outside `GITHUB_WORKSPACE`; the CHANGELOG entry calls this out.

## Test plan

- [x] `npm test` — 268 tests pass
- [x] `npm run typecheck` — clean
- [x] `npm run build` — clean
- [ ] Reviewers: confirm the new behavior aligns with expectations for self-hosted runners that intentionally serve manifests from outside the workspace


CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # Changelog
 
+## Unreleased
```
Tatsinnit: 💡 ❤️  General note please: I see @davidgamero or @bosesuneha  might know the right pattern followed in this repo for this release, we can always make this part of "Added" post this review. Jsut a gentle point out. Thank heaps.

gambtho: removed for now to allow faster merge

Tatsinnit (approved):

davidgamero (approved):
lgtm

Foxboron:
This breaks usage with `k8s-bake` as there is no way to tell the action step to output the files inside the workspace.

PR #542, opened 2026-06-30
fix: support GitHub Enterprise Server domains in annotation URLs

### LINKED ISSUE #201: Fix annotations links using github.com even for GHES
There are few locations were github.com is hardcoded, making it incompatible with github enterprise servers that are hosted on other domains.
From what I understood from these code lines, they are only cosmetic (sent as annotations in deployments).

https://github.com/Azure/k8s-deploy/blob/507f2d4fc713cdc92f96880e350464687ce2e960/src/utilities/workflowAnnotationUtils.ts#L24
https://github.com/Azure/k8s-deploy/blob/ca8d2604acf262265b6640fc4de4e060effbb555/src/utilities/githubUtils.ts#L47

There is `GITHUB_SERVER_URL` env var and [some others](https://docs.github.com/en/enterprise-server@2.22/actions/learn-github-actions/environment-variables) to get the server domain.

ejaifeobuks:
Replace hardcoded https://github.com in the deployment annotation runUri and getNormalizedPath blob links with GITHUB_SERVER_URL, falling back to https://github.com when the var is unset. This makes the generated links correct on GitHub Enterprise Server instances hosted on custom domains.

This PR fixes issue 
This pull request improves how URLs are constructed for GitHub resources by ensuring that the correct GitHub server URL is used, especially in environments where GitHub Enterprise or custom domains are in use. The changes make the code more flexible and robust for different deployment scenarios.

**URL Construction Improvements:**

* Updated `getNormalizedPath` in `src/utilities/githubUtils.ts` to use `GITHUB_SERVER_URL` (defaulting to `https://github.com`) when constructing blob URLs, instead of hardcoding the domain.
* Updated `getWorkflowAnnotations` in `src/utilities/workflowAnnotationUtils.ts` to use `GITHUB_SERVER_URL` (defaulting to `https://github.com`) when constructing workflow run URLs, instead of hardcoding the domain.
* This PR fixes issue #201

### DIFF
--- a/src/utilities/githubUtils.ts
+++ b/src/utilities/githubUtils.ts
@@ -44,7 +44,8 @@ export function normalizeWorkflowStrLabel(workflowName: string): string {
 export function getNormalizedPath(pathValue: string) {
    if (!isHttpUrl(pathValue)) {
       //if it is not an http url then convert to link from current repo and commit
-      return `https://github.com/${process.env.GITHUB_REPOSITORY}/blob/${process.env.GITHUB_SHA}/${pathValue}`
+      const serverUrl = process.env.GITHUB_SERVER_URL || 'https://github.com'
+      return `${serverUrl}/${process.env.GITHUB_REPOSITORY}/blob/${process.env.GITHUB_SHA}/${pathValue}`
    }
    return pathValue
 }
--- a/src/utilities/workflowAnnotationUtils.ts
+++ b/src/utilities/workflowAnnotationUtils.ts
@@ -16,7 +16,11 @@ export function getWorkflowAnnotations(
       workflowFileName: workflowFilePath.replace('.github/workflows/', ''),
       jobName: process.env.GITHUB_JOB,
       createdBy: process.env.GITHUB_ACTOR,
-      runUri: `https://github.com/${process.env.GITHUB_REPOSITORY}/actions/runs/${process.env.GITHUB_RUN_ID}`,
+      runUri: `${
+         process.env.GITHUB_SERVER_URL || 'https://github.com'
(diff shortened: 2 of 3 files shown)


src/utilities/workflowAnnotationUtils.ts
```diff
@@ -16,7 +16,11 @@ export function getWorkflowAnnotations(
       workflowFileName: workflowFilePath.replace('.github/workflows/', ''),
       jobName: process.env.GITHUB_JOB,
       createdBy: process.env.GITHUB_ACTOR,
-      runUri: `https://github.com/${process.env.GITHUB_REPOSITORY}/actions/runs/${process.env.GITHUB_RUN_ID}`,
+      runUri: `${
+         process.env.GITHUB_SERVER_URL || 'https://github.com'
```
bosesuneha: can you extract this into a const instead of having it inline? it's just easier to read.

src/utilities/githubUtils.ts
```diff
@@ -44,7 +44,8 @@ export function normalizeWorkflowStrLabel(workflowName: string): string {
 export function getNormalizedPath(pathValue: string) {
    if (!isHttpUrl(pathValue)) {
       //if it is not an http url then convert to link from current repo and commit
-      return `https://github.com/${process.env.GITHUB_REPOSITORY}/blob/${process.env.GITHUB_SHA}/${pathValue}`
+      const serverUrl = process.env.GITHUB_SERVER_URL || 'https://github.com'
```
bosesuneha, line 47: can we test case for this?

ejaifeobuks: Sure! I will add one in the next commit

PR #560, opened 2026-08-26
fix(security): command injection via az aks command invoke

davidgamero:
PrivateKubectl.execute flattened its argv array with args.join(' ') and passed the result to `az aks command invoke --command`. 

The run-command pod uses its own high-privilege service account, so this escapes whatever RBAC the workflow's kubeconfig was scoped to.

Values reaching the command string:

  - manifest metadata.name, via the temp filename passed to -f. getNewTempManifestFileName applies path.basename(), which strips path separators but not ';', '|', '$(...)', backticks or spaces.
  - the namespace input, via --namespace. Workflows commonly build this from github.head_ref, and git permits ';', '$', '(', ')', '`', '|' and '&' in branch names.
  - the workflow annotation payload, which is unescaped JSON containing spaces and double quotes plus GITHUB_WORKFLOW and the branch name.

Fix: rewrite filename arguments positionally on the argv array, then POSIX-quote every element before joining. Quoting is only possible once the filename rewriting no longer needs to re-parse a flattened string, so the two changes are inseparable.

This also removes the minimist re-parse and the '-f ' substring test, which mis-handled any value containing a space and could rewrite unrelated arguments.

Also honours the caller's `silent` flag: command output was previously written to the debug log unconditionally, including for callers such as getAllPods() that explicitly request silence.

Tests verify the fix by executing the generated command in a real POSIX shell with kubectl stubbed to dump its argv, then asserting it reconstructs the original array exactly, for payloads carried in annotations, resource names, namespaces and filenames. 

Also removes minimist as it's unused dependency

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,12 +15,10 @@
             "@actions/tool-cache": "4.0.0",
             "@octokit/core": "^7.0.6",
             "@octokit/plugin-retry": "^8.1.0",
-            "js-yaml": "5.2.2",
-            "minimist": "^1.2.8"
+            "js-yaml": "5.2.2"
          },
          "devDependencies": {
             "@types/js-yaml": "^4.0.9",
-            "@types/minimist": "^1.2.5",
             "@types/node": "^26.1.1",
             "esbuild": "^0.28",
             "husky": "^9.1.7",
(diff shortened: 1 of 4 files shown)


bosesuneha (approved):
lgtm

